PR #1, opened 2016-05-19
Adding the 1-click cross-validation example

mmerce:
Please, mind testing the new scripts before merging.


whizzmler:
merged! i've added a further commit that groups them together in a package (although they're still usable as individual scripts)

PR #2, opened 2016-05-25
gradient-boosting: shorter resource names

jaor:


charleslparker:
Did you run this to make sure it actually works?  I've just been finicky about names in places.

Also I saw the thread where you mention that this is a hacky fix and of course I agree completely; anyone who is creating resources from one another in a looping fashion will run into this in pretty short order.

jaor:
yes, i ran it and seems to be working... but as a reviewer you should
run it too instead of trusting me :) with iris it doesn't take too long.

PR #7, opened 2016-06-08
Added empty list as default value for train-exc input parameter

hangarr:
Added an empty list value as the default value for the train-exc parameter as required with scripts in the BigML Dashboard.


jaor:
looks good, thanks!

PR #11, opened 2016-07-12
Adds script for global field importance for clusters

akashenfelter:
A script to determine which input fields are globally most important when differentiating between clusters. 

@jaor, would it be appropriate to put in a PR to add `merge-with` to the WhizzML standard library?

### DIFF
--- a/cluster-classification/readme.md
+++ b/cluster-classification/readme.md
@@ -1,0 +1,11 @@
+# Global Field Importance for Clusters
+
+This script determines which input fields are most important for
+differentiating between clusters. Given a cluster id, it:
+
+- Using batchcentroid, extends the dataset the cluster was
+  built on with the centroid id each instance is a member of.
+- Builds an ensemble (random forest) on that centroid id field.
+- Averages the importance of each input field across all models
+  in the ensemble.
+- Returns a sorted list of input fields by importance.
--- a/cluster-classification/script.whizzml
+++ b/cluster-classification/script.whizzml
@@ -9,8 +9,8 @@
 
 (define (label-by-cluster cl-id) 
   (let (cl (fetch cl-id)
-        cl-ds (get cl "dataset")
-        cl-fields (get cl "input_fields"))
+        cl-ds (cl "dataset")
+        cl-fields (cl "input_fields"))
     (create-and-wait-batchcentroid
      {"cluster" cl-id 
       "dataset" cl-ds 
@@ -40,10 +40,10 @@
 ;; field ids and average importances
 
 (define (make-importance-map en-id ids)
-  (let (models (get-in (fetch en-id) ["models"])
+  (let (models ((fetch en-id) ["models"])
         len-models (count models)
         imp-lst (map (lambda (md-id) 
-                       (get-in (fetch md-id) ["model" "importance"])) 
+                       ((fetch md-id) ["model" "importance"])) 
                      models)
         imp-map (map list-to-map imp-lst)
         imp-sum (reduce (lambda (x y) (merge-with + x y)) {} imp-map))
@@ -63,11 +63,11 @@
         labeled (bc "output_dataset_resource")
         new-ids (butlast ((fetch labeled) "input_fields"))
         cl-en (create-and-wait-ensemble labeled {"randomize" true})
-        ds-fields (get (fetch labeled) "fields")
+        ds-fields ((fetch labeled) "fields")
         imp-map (make-importance-map cl-en new-ids)
         name-map (make-map new-ids 
                            (map (lambda (x) 
-                                  (get-in ds-fields [x "name"])) 
+                                  (ds-fields [x "name"])) 
                                 new-ids))
         id-map (make-map new-ids old-ids)       
         total-map (map (lambda (x) 
--- a/cluster-classification/script.whizzml
+++ b/cluster-classification/script.whizzml
@@ -1,0 +1,84 @@
+;; Given a cluster, creates a batch centroid and uses the output
+;; dataset where each instance is labeled by centroid to create an
+;; ensemble (random forest), and returns the importance of each field
+;; averaged over the ensemble.
+
+;; Given a cluster, returns a batchcentroid id created from that
+;; cluster whose output dataset is the original cluster dataset
+;; appended with the centroids the instances are part of.
+
+(define (label-by-cluster cl-id) 
+  (let (cl (fetch cl-id)
+        cl-ds (get cl "dataset")
+        cl-fields (get cl "input_fields"))
+    (create-and-wait-batchcentroid
+     {"cluster" cl-id 
+      "dataset" cl-ds 
+      "output_dataset" true
+      "output_fields" cl-fields})))
+
+;; Given a list of pairs, creates the equivalent map
+
+(define (list-to-map nested-list)
+  (let (key-list (map head nested-list)
+        value-list (map (lambda (x) (last x)) nested-list))
+    (make-map key-list value-list)))
+
+;; Given a function of two values and two maps, returns a new map
+;; where if a key appears in only one of the original maps its value
+;; is unchanged, but if it appears in both its value is the function
+;; of the two original values.  
+
+(define (merge-with fn map1 map2) 
+  (let (subset-keys (filter (lambda (x) (contains? map1 x)) (keys map2))
+        subset-values (map (lambda (x) 
+                             (fn (get map1 x) (get map2 x))) subset-keys) 
+        subset (make-map subset-keys subset-values)) 
+    (merge map1 (merge map2 subset))))
+
+;; Given a ensemble id and a list of field ids, returns a map of those
+;; field ids and average importances
+
+(define (make-importance-map en-id ids)
+  (let (models (get-in (fetch en-id) ["models"])
+        len-models (count models)
+        imp-lst (map (lambda (md-id) 
+                       (get-in (fetch md-id) ["model" "importance"])) 
+                     models)
+        imp-map (map list-to-map imp-lst)
+        imp-sum (reduce (lambda (x y) (merge-with + x y)) {} imp-map))
+    (make-map ids (map (lambda (x) 
+                         (if (contains? imp-sum x) 
+                           (/ (get imp-sum x) len-models) 0)) 
+                       ids))))
+
+;; Given a cluster-id, builds a random forest on centroid membership,
+;; and returns a paired list of field importance averaged across all
+;; models in the forest
+
+(define (cluster-classification cl-id)
+  (let (bc-id (label-by-cluster cl-id)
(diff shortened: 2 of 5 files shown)


cluster-classification/script.whizzml
```diff
@@ -0,0 +1,70 @@
...
+;; dataset where each instance is labeled by centroid to create an
+;; ensemble (random forest), and returns the importance of each field
+;; averaged over the ensemble.
+
+;; Given a cluster, returns a batchcentroid id created from that
+;; cluster whose output dataset is the original cluster dataset
+;; appended with the centroids the instances are part of.
+
+(define (label-by-cluster cl-id) 
+  (let (cl-ds (get-in (fetch cl-id) ["dataset"])
```
jaor: we're just releasing a new version of whizzml where maps and lists can be used as procedures, so that the idiomatic way of writing the above will become:

```
((fetch cl-id) "dataset")
```

it's fine if you prefer to send a PR afterwards modifying that, but you can also wait a day and fix it in this PR.

If we go with the current version, the currently idiomatic way would be using `get`:

```
(get (fetch cl-id) "dataset")
```

cluster-classification/script.whizzml
```diff
@@ -0,0 +1,70 @@
...
+;; ensemble (random forest), and returns the importance of each field
+;; averaged over the ensemble.
+
+;; Given a cluster, returns a batchcentroid id created from that
+;; cluster whose output dataset is the original cluster dataset
+;; appended with the centroids the instances are part of.
+
+(define (label-by-cluster cl-id) 
+  (let (cl-ds (get-in (fetch cl-id) ["dataset"]) 
+        cl-fields (get-in (fetch cl-id) ["input_fields"]))
```
jaor: we are actually caching the result of calling `fetch`, but it's nonetheless a relatively heavy op:

```
(let (cl (fetch cl-id)
      cl-ds (cl "dataset")
      cl-fields (cl "fields"))
  ...)
```

cluster-classification/script.whizzml
```diff
@@ -0,0 +1,70 @@
...
+    (create-and-wait-batchcentroid 
+     {"cluster" cl-id 
+      "dataset" cl-ds 
+      "output_dataset" true
+      "output_fields" cl-fields})))
+
+;; Given a list of pairs, creates the equivalent map
+(define (list-to-map nested-list)
+  (let (key-list (map head nested-list)
+        value-list (map (lambda (x) (nth x 1)) nested-list))
```
jaor: in the new-style whizzml the lambda body will become

```
(lambda (x) (x 1))
```

ashenfad: Jumping in - but since these lists have just two values, you can drop the lambda fn altogether in the new WhizzML and just use `last`.

cluster-classification/script.whizzml
```diff
@@ -0,0 +1,70 @@
...
+;; Given a function of two values and two maps, returns a new map
+;; where if a key appears in only one of the original maps its value
+;; is unchanged, but if it appears in both its value is the function
+;; of the two original values.  
+(define (merge-with fn map1 map2) 
+  (let (subset-keys (filter (lambda (x) (contains? map1 x)) (keys map2))
+        subset-values (map (lambda (x) (fn (get map1 x) (get map2 x))) 
+                           subset-keys) 
+        subset (make-map subset-keys subset-values)) 
+    (merge map1 (merge map2 subset))))
```
jaor, line 37: This is a very nice implementation! Here's an alternative that i
suspect is more efficient, at the cost of some elegance:

```
(let (map2-ex (iterate (r map2 k (keys map2))
                (if (contains? map1 k)
                    (assoc r k (fn (map1 k) (map2 k)))
                    r)))
  (merge map1 map2-ex))
```

If you submit a PR to incorporate merge-with to the stdlib, we might
go for performance, but if you find your implementation clearer and a
couple measures with the REPL's ,time function don't show it's very
slow, i'll be fine with your implementation too.

cluster-classification/script.whizzml
```diff
@@ -0,0 +1,70 @@
...
+
+;; Given a cluster-id, builds a random forest on centroid membership,
+;; and returns a paired list of field importance averaged across all
+;; models in the forest
+
+(define (cluster-classification cl-id)
+  (let (bc-id (label-by-cluster cl-id)
+        labeled (get-in (fetch bc-id) ["output_dataset_resource"])
+        _ (wait labeled)
+        cl-en (create-and-wait-ensemble {"dataset" labeled
```
jaor: in the new version of whizzml, waits happen automatically for you and
create functions accept ids too, so you'll be able to write the above as:

```
(let (bc-id (label-by-cluster cl-id)
      labeled ((fetch bc-id) "output_dataset")
      cl-en (create-ensemble labeled {"randomize" true})
```

cluster-classification/script.whizzml
```diff
@@ -0,0 +1,70 @@
...
+        models (get-in (fetch cl-en) ["models"])
+        len-models (count models)
+        imp-lst (map (lambda (md-id) 
+                       (get-in (fetch md-id) ["model" "importance"])) 
+                     models)
+        imp-map (map list-to-map imp-lst)
+        imp-sum (reduce (lambda (x y) (merge-with + x y)) {} imp-map))
+    (delete labeled)
+    (delete bc-id)
+    (delete cl-en)
```
jaor: you can write that more succintly as:

```
(delete* [labeled bc-id cl-en])
```

cluster-classification/script.whizzml
```diff
@@ -0,0 +1,70 @@
...
+    (map (lambda (x) (list (head x) (/ (nth x 1) len-models))) 
+         imp-sum)))
```
jaor: hmm, is this corret? `imp-sum` is a map, isn't it?

jaor:
looks pretty good, just a few comments.  most of them are about adapting the code to whizzml 0.17, which will be released today or tomorrow... we can either wait for it, or merge this in old-style and then you submit a new PR to "modernize" it.

also, it'd be very nice if you could write a test (using bigmler as the driver) similar to those in `cross-validation` or `boruta`... they're essentially a bash script that uses bigmler to run the script and then check that the result make sense.

cluster-classification/script.whizzml
jaor: yes, even nicer!

cluster-classification/script.whizzml
akashenfelter: Waiting and fixing things up for whizzml 0.17 sounds like a good idea. I'll make all those changes and wait for the next version.

cluster-classification/script.whizzml
akashenfelter: Excellent, this is how to get multiple elements from a single fetch. I didn't like how I was doing it at all.

cluster-classification/script.whizzml
akashenfelter, line 37: I'll do some tests with the REPL's time function. I do think my version is more transparent, but of course I wouldn't want a noticeable hit to performance.

cluster-classification/script.whizzml
akashenfelter: `imp-sum` is a map. Here I'm doing two things at once: converting `imp-sum` to a list of pairs (which is how importance is stored in the model object) and normalizing the summed importance value. Is it abusing `map` to use it to convert maps to lists? Should I pull this apart to make it more clear what's going on?

akashenfelter:
Will do! I'll add the tests using `cross-validation` and `boruta` as examples to guide me. Thanks!

cluster-classification/script.whizzml
jaor: hmmm, this is funny... the ability to map over maps (and their automatic
transformation to lists of pairs) is actually an accidental feature of
whizzml! i didn't implement it on purpose (and that's why it's
undocumented), and i'm not sure whether it'll work consistently.  but
since it's already working for you and you've found it useful, i'll just
write tests for it and document it in whizzml. thanks!

whizzmler: although i'm having second thoughts on this... if we're going to allow
map over maps, it seems more natural to implement it so that the
function takes two arguments, key and value, so that you could write
your function:

```
 (map (lambda (k v) [k (/ v len-models)]) imp-sum)
```

and for multiple maps

```
 (map (lambda (k0 v0 k1 v1) ...) m0 m2)
```

and we can even mix maps and lists:

```
 (map (lambda (k0 v0 l k1 v1) ...) m0 ls m2)
```

whizzmler: and, regardless of what we finally do with map over maps, yours right
now is easier to write and read like this (this already works):

```
(map (lambda (h t) [h (/ t len-models)]) (keys imp-sum) (values imp-sum))
```

since map can take multiple lists (and [x y] and (list x y) are equivalent).

ashenfad: What about adding Clojure-like bindings in future versions of WhizzML?  So the lambda fn would still take a single var, but the binding gets the key and value:

``` clojure
(map (lambda ((k v)) ... ) some-hashmap)
```

jaor: that's hard to implement, and the other solutions work good enough to
not completely justify that effort

akashenfelter:
Nits picked, rebased and squashed. 

I also made some changes to the output field; it now contains the field name, for much greater human readability. It was a little difficult because the batchcentroid field ids aren't the same as the dataset field ids. Is this something that could be changed?

The test script is coming in a future pull request, after I have a machine that can more easily run bash scripts. This pull request has been sitting so long, I thought it worth committing and closing even though it's not as nice as I'd like.

cluster-classification/script.whizzml
```diff
@@ -0,0 +1,84 @@
...
+;; ensemble (random forest), and returns the importance of each field
+;; averaged over the ensemble.
+
+;; Given a cluster, returns a batchcentroid id created from that
+;; cluster whose output dataset is the original cluster dataset
+;; appended with the centroids the instances are part of.
+
+(define (label-by-cluster cl-id) 
+  (let (cl (fetch cl-id)
+        cl-ds (get cl "dataset")
```
jaor: we could remove all get and get-in...

cluster-classification/readme.md
```diff
@@ -0,0 +1,11 @@
+# Global Field Importance for Clusters
```
jaor, line 1: perhaps we can also include this one in the top-level metadata.json as a component? and surely to the top-level readme.md in the examples list?

jaor:
looks pretty good: just a couple nits.

akashenfelter:
New nits picked! Bash shell comes to Windows 10 tomorrow, so that test script will be here very soon.

jaor:
looks good.  looking forward to that test script! merged, thanks a lot!

PR #12, opened 2016-07-20
Adding the boruta feature selection script

mmerce:


boruta/1-click-boruta/script.whizzml
```diff
@@ -0,0 +1,386 @@
...
+;; choosable-objective-ids
+;;
+;; List of IDs of the fields in the dataset that can be chosen as objective
+;; field.
+;;
+;; Inputs:
+;;  fields: (map) Fields structure
+;; Output: (list) list of field IDs
+(define (choosable-objective-ids fields)
+  (let (field-val (lambda (fid k) (get-in fields [fid k]))
```
jaor: `(fields [fid k])` ... and all `get-in` and `get` elsewhere can be changed too...

boruta/1-click-boruta/script.whizzml
```diff
@@ -0,0 +1,386 @@
...
+  (let (ds (fetch dataset-id)
+        ;; original fields are stored for later use
+        ori-fields (get ds "fields" {})
+        field-ids (keys ori-fields)
+        f-names (map (lambda (x) (get-in ori-fields [x "name"] false))
+                     field-ids)
+        init-rank (reduce (lambda (x y) (assoc x (get-in ori-fields
+                                                         [y "name"]
+                                                         false) 0))
+                          {}
```
jaor: it's more idiomatic to cut the lambda this way:

```
(reduce (lambda (x y)
          (assoc x (ori-fields [y "name"] false) 0))
        {}
        field-ids)
```

although this is even shorter with iterate:

```
(iterate (r {} id field-ids)
  (assoc r (ori-fields [id "name"] false) 0))
```

Also, you want to use `false` as a key in the result when an id
doesn't exist instead of throwing and error? (just double-checking)

boruta/1-click-boruta/script.whizzml
```diff
@@ -0,0 +1,386 @@
...
+          ;; Loop body: Extends the information system by adding copies of all
+          ;; variables (the information system is always extended by at least
+          ;; 5 shadow attributes, even if the number of attributes in the
+          ;; original set is lower than 5)
+          ;; Shuffle the added attributes to remove their correlations with
+          ;; the response.
+          (let (sh-ds-info (shadow-ds ds-id fields objective)
+                sh-ds-id (head sh-ds-info)
+                sh-ids (head (tail sh-ds-info))
+                fields (nth sh-ds-info 2)
```
jaor: for lists, `nth` is not needed any more (applicable to full script), and, for clarity,  i'd write the three values above homogeneously as:

```
sh-ds-id (sh-ds-info 0)
sh-ids (sh-ds-info 1)
fields (sh-ds-info 2)
```

boruta/1-click-boruta/script.whizzml
```diff
@@ -0,0 +1,386 @@
...
+                to-names (lambda (x) (get-in fields [x "name"] false))
```
jaor: i was confused by the name of this function, which sounds as if it returned a list of names... perhaps `to-name`?

boruta/1-click-boruta/script.whizzml
```diff
@@ -0,0 +1,386 @@
...
+;; full
+;; Checks whether all the fields are marked as important or unimportant
+;;
+;; Inputs:
+;;   f-rank: (map) Map from field ID to imporance flag
+;;
+;; Output: (boolean) True if all the fields are classified
+;;
+(define (full f-rank)
+  (let (undetermined (filter (lambda (x) (= x 0)) (values f-rank)))
```
jaor: we have `zero?`: `(undetermined (filter zero? (values f-rank)))`

boruta/1-click-boruta/script.whizzml
```diff
@@ -0,0 +1,386 @@
...
+;;
+;; Inputs:
+;;   fields: (map) Fields information structure
+;;   f-rank: (map) Map from field ID to importance flag
+;;
+;; Output: (map) Fields information structure without unimportant fields
+;;
+(define (remove-unimportant fields f-rank)
+  (iterate (acc fields id (keys f-rank))
+    (if (< (get f-rank id 1) 0)
```
jaor: `(negative? (f-rank id 0))`

boruta/1-click-boruta/script.whizzml
```diff
@@ -0,0 +1,386 @@
...
+;;   f-rank: (map) Map from field ID to importance flag
+;;   fields: (map) Fields information structure
+;;   f-names: (list) Fields names
+;;   objective: (string) objective field name
+;;
+;; Output:  (map) Map from categories (important, unimportant or undecided) to
+;;                list of field names
+;;
+(define (group-importance f-rank fields f-names objective)
+  (let (to-names (lambda (x) (get-in fields [x "name"] false))
```
jaor: same comment as before for the name of the function.

boruta/1-click-boruta/script.whizzml
```diff
@@ -0,0 +1,386 @@
...
+        important (map to-names
+                       (keys (filter-map (lambda(x y) (> y 0))
+                                         f-rank)))
```
jaor: an alternative in one pass, but yours is fine if you like it:

```
important (iterate (r [] k (keys f-rank))
            (if (positive? (f-rank k)) (append r k) k))
```

boruta/1-click-boruta/script.whizzml
```diff
@@ -0,0 +1,386 @@
...
+        important (append important objective)
+        undecided (map to-names
+                       (keys (filter-map (lambda(x y) (= y 0))
+                                         f-rank)))
```
jaor: and similarly this one with `s/positive?/zero?`

boruta/1-click-boruta/script.whizzml
```diff
@@ -0,0 +1,386 @@
...
+        unimportant (filter (lambda (x) (not (or (member? x important)
+                                                 (member? x undecided))))
+                            f-names))
```
jaor: it's more idiomatic, if a nit, to break the lambda after the `(x)`

boruta/1-click-boruta/script.whizzml
```diff
@@ -0,0 +1,386 @@
...
+;; Filters a map by applying the corresponding filter function to its keys
+;;
+;; Inputs:
+;;   fn: (function) function to be applied to the map keys and/or values
+;;   map: (map) Map to be filtered
+;;
+;; Output: (map) Filtered map
+(define (filter-map fn map)
+    (iterate (out-map {} key (keys map))
+      (if (fn key (get map key false))
```
jaor: nit, but, in principle, it's impossible that `(get map key)` will fail here.

also, i recommend not using `map` as the name of a local variable, because it shadows the syntactic form `map`, and you cannot use `(map key)` because `map` is syntax.

boruta/1-click-boruta/script.whizzml
```diff
@@ -0,0 +1,386 @@
...
+        input-fields (cons obj-id f-ids)
+        f-ids (repeat-fields f-ids 5)
+        f-names (map (lambda (x) (str "shadow " (get f-id-names x false)))
+                     f-ids)
+        flatline-fs (map (lambda (x) [(flatline
+                                        "(weighted-random-value {{x}})")
+                                      (str "shadow " (get f-id-names x))])
+                         f-ids)
+        new-fields (lambda (x) (assoc {} "field" (head x)
+                                         "name" (head (tail x))))
```
jaor: i think it's much more clear to define `new-fields` with a single list comprehension, without the intermediate `flatline-fs`:

```
new-fields (for (id f-ids)
             {"field" (flatline "(weighted-random-value {{id}})")
              "name" (str "shadow " (f-id-names id))})
```

jaor: or, since you have `f-names` you can also use a map, again without `flatline-fs`:

```
new-fields (map (lambda (id name) {"field" id "name" name})
                f-ids
                f-names)
```

boruta/1-click-boruta/script.whizzml
```diff
@@ -0,0 +1,386 @@
...
+                                           "input_fields"
+                                           input-fields
+                                           "new_fields"
+                                           (map new-fields flatline-fs)})
+        sh-ds (fetch sh-ds-id)
+        sh-fields (get sh-ds "fields" false)
+        sh-f-ids (filter (lambda (x) (member? (get-in sh-fields
+                                                      [x "name"]
+                                                      false)
+                                              f-names))
```
jaor: `lambda` body in new line

boruta/1-click-boruta/script.whizzml
```diff
@@ -0,0 +1,386 @@
...
+
+(define (get-importances ensemble)
+  (let (dist (get ensemble "distributions" false)
+        imps (map (lambda (x) (get x "importance" false)) dist)
+        m-num (count imps))
+    (iterate (acc {} m-imps imps)
+      (iterate (m-acc acc imp m-imps)
+        (assoc+ m-acc
+                (head imp)
+                (/ (head (tail imp)) m-num))))))
```
jaor: here again, i'd use `(imp 0)` and `(/ (imp 1) m-num)`

boruta/boruta/readme.md
```diff
@@ -0,0 +1,70 @@
+# Boruta algorithm
```
jaor: we could just say here `see [../1-click-boruta/readme.md](blah blah)` no? or even do a `ln -s`... is there any difference between the two?

boruta/boruta/script.whizzml
```diff
@@ -0,0 +1,362 @@
+;; This will eventually be packed into a library
```
jaor: i guess all the comments above apply to this file too.  do we really need all this duplication? why not include only the 1-click, and we'll refactor to library when we can?

boruta/readme.md
```diff
@@ -0,0 +1,30 @@
+# Boruta Feature Selection
+
+In this package you'll find two scripts implementing Boruta Feature Selection:
+
+- [Boruta 1-click Feature Selection](./1-click-boruta)
+  Starting from a dataset, creates a new one removing the unimportant features
+  according to the Boruta Feature Selection algorithm.
+- [Boruta Feature Selection](./boruta)  General Boruta Feature Selection
+  algorithm to rank fields from a dataset into three categories: important,
+  unimportant and undecided.
```
jaor: what's the difference with the previous one? i would have the 1-click-boruta, and add a library when we can...

jaor:
some comments in!

boruta/1-click-boruta/script.whizzml
mmerce: wouldn't it be?

important (iterate (r [] k (keys f-rank))
            (if (positive? (f-rank k)) (append r k) r))

and then I'd lack the tranformation from ids to names...

important (iterate (r [] k (keys f-rank))
            (if (positive? (f-rank k)) (append r (to-name k)) r))

boruta/1-click-boruta/script.whizzml
mmerce: right, I just bulk-changed every get to avoid raising errors.

what name would you like for the map argument? map-struct?

boruta/1-click-boruta/script.whizzml
mmerce: you mentioned first option is easier to adopt for non-lisp programmers, so i'd go for this one. It's good to know that map can act on more than one list, though!

boruta/readme.md
mmerce: Well, the output of the original boruta script is just a map of "important", "unimportant" and "undecided" fields. The 1-click creates a new dataset removing the fields marked as unimportant and gives defaults for every argument except the datase ID, so objective field is the one in the original dataset, the maximum number of runs is 10 and the epsilon arount the maximum importance of the shadow fields is 0.01

mmerce: I wrote both just following the cross-validation example (I think they are published separately in the gallery) but if you think it's not worth the effort right now, I can remove the boruta folder and leave just the 1-click.

boruta/1-click-boruta/script.whizzml
jaor: a-map

mmerce: the-larch! :-)

mmerce:
I believe this is what you asked for

mmerce:
just a note to @ashenfad and @a.k.ashenfad to let them know that the script has gone through code review now.

PR #19, opened 2016-09-07
Added test.sh to cluster-classification

akashenfelter:


jaor:
looks good and the test passes: merged, thanks!

PR #22, opened 2016-09-23
Auto-complete Dataset

charleslparker:
This is a little script that takes a dataset that may have missing data, trains a model for each missing field, and uses the model to fill in the missing data.  It also inserts empty strings for missing text/items fields and removes non-preferred fields in the bargain.  Finally, for each missing field that it had to "complete", it adds a column indicating whether or not the field was missing in the original dataset.

This may be useful for the upcoming Laminar, as DBNs don't deal with missing data in a natural way.  Also for logistic regression where we don't feel like our built-in options for imputation are enough (/cc @cheesinglee).

This was inspired by the talks that Poul and I gave in Valencia, as well as a conversation I had afterwards with someone who asked for it specifically.

@jaor - Is there a reason that Poul and FM aren't in this group?  I'd like to cc them but I guess I can't unless they are in the whizzml "organization".

### DIFF
--- a/repair-missing/metadata.json
+++ b/repair-missing/metadata.json
@@ -1,25 +1,37 @@
 {
   "name": "Auto-complete Missing Fields",
-  "description": "Use dataset information to fill in missing values for numeric or categorical fields which have at least one value missing.  For numeric fields, we learn a random forest to predict a probable field value given the information that is present.  For categorical fields we add an explict 'missing' category to the field.  In both cases, we also add a binary column indicating whether the value was missing in the original dataset or not.",
+  "description": "This script fills in missing values for all columns in a given dataset.  For numeric and categorical fields, this script uses the input fields of the dataset to learn models that predict the missing values.  For text and items fields, missing values are replaced by the empty string.  We also add a binary column indicating whether the value was missing in the original dataset or not.  Non-preferred fields are removed.  See the readme for more information.",
   "kind": "script",
   "source_code": "script.whizzml",
   "inputs": [
       {
           "name": "dataset-id",
           "type": "dataset-id",
-          "description": "Dataset to be completed"
+          "description": "Dataset to be Completed"
       },
       {
-          "name": "objective-field-id",
+          "name": "objective-field",
           "type": "string",
-          "description": "Objective field of the input dataset, or null to choose the default objective"
+          "default": "",
+          "description": "Objective field of the input dataset, or the empty string to choose the default objective.  If use-objective is true, this field is excluded from the missing field models and placed at the end of the output dataset."
+      }
+      {
+          "name": "use-objective",
+          "type": "boolean",
+          "default": true,
+          "description": "Whether or not to use an objective field"
       }
   ],
   "outputs": [
       {
-          "name": "completed-dataset",
+          "name": "repaired-dataset",
           "type": "dataset-id",
           "description": "Dataset with completed columns added"
+      }
+      {
+          "name": "processor",
+          "type": "map",
+          "description": "A collection of information about the repair process, including the models learned to replace missing data.  Can be to complete future datasets of the same format."
       }
   ]
 }
--- a/repair-missing/metadata.json
+++ b/repair-missing/metadata.json
@@ -31,7 +31,7 @@
       {
           "name": "processor",
           "type": "map",
-          "description": "A collection of information about the repair process, including the models learned to replace missing data.  Can be to complete future datasets of the same format."
+          "description": "A collection of information about the repair process, including the models learned to replace missing data.  Can be used to complete future datasets of the same format."
       }
   ]
 }
--- a/repair-missing/metadata.json
+++ b/repair-missing/metadata.json
@@ -1,0 +1,25 @@
+{
+  "name": "Auto-complete Missing Fields",
+  "description": "Use dataset information to fill in missing values for numeric or categorical fields which have at least one value missing.  For numeric fields, we learn a random forest to predict a probable field value given the information that is present.  For categorical fields we add an explict 'missing' category to the field.  In both cases, we also add a binary column indicating whether the value was missing in the original dataset or not.",
+  "kind": "script",
+  "source_code": "script.whizzml",
+  "inputs": [
+      {
+          "name": "dataset-id",
+          "type": "dataset-id",
+          "description": "Dataset to be completed"
+      },
+      {
+          "name": "objective-field-id",
+          "type": "string",
+          "description": "Objective field of the input dataset, or null to choose the default objective"
+      }
+  ],
+  "outputs": [
+      {
+          "name": "completed-dataset",
+          "type": "dataset-id",
+          "description": "Dataset with completed columns added"
+      }
+  ]
+}
--- a/repair-missing/metadata.json
+++ b/repair-missing/metadata.json
@@ -14,7 +14,7 @@
           "type": "string",
           "default": "",
           "description": "Objective field of the input dataset, or the empty string to choose the default objective.  If use-objective is true, this field is excluded from the missing field models and placed at the end of the output dataset."
-      }
+      },
       {
           "name": "use-objective",
           "type": "boolean",
@@ -27,7 +27,7 @@
           "name": "repaired-dataset",
           "type": "dataset-id",
           "description": "Dataset with completed columns added"
-      }
+      },
       {
           "name": "processor",
           "type": "map",
--- a/repair-missing/readme.md
+++ b/repair-missing/readme.md
@@ -1,0 +1,21 @@
+# Missing Field Completion
+
+## Tests
+
+The `test` directory contains a shell script named `test.sh`
+which uses `BigMLer` to perform a basic test of the WhizzML code. To run the
+test:
+
+- BigMLer must be installed. For instructions to install BigMLer please refer
+to the [BigMLer documentation](http://bigmler.readthedocs.io/en/latest/#bigmler-installation).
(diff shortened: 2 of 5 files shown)


jaor:
> @jaor - Is there a reason that Poul and FM aren't in this group? I'd
> like to cc them but I guess I can't unless they are in the whizzml
> "organization".

@aficionado hasn't accepted his invitation yet (sent months ago).  i
just sent one to @petersen-poul.

whizzmler:
> @jaor https://github.com/jaor - Is there a reason that Poul and FM
> aren't in this group? I'd like to cc them but I guess I can't unless they
> are in the whizzml "organization".

  ? I got this...

-poul

charleslparker:
Poul - it looks like you got it as the user @whizzmler, but @petersen-poul doesn't belong to the group.  In any case, I thought not belonging to the group meant that @-mentions wouldn't parse, but I guess they just don't auto-complete and are still communicated.

repair-missing/metadata.json
```diff
@@ -0,0 +1,37 @@
...
+  "outputs": [
+      {
+          "name": "repaired-dataset",
+          "type": "dataset-id",
+          "description": "Dataset with completed columns added"
+      },
+      {
+          "name": "processor",
+          "type": "map",
+          "description": "A collection of information about the repair process, including the models learned to replace missing data.  Can be to complete future datasets of the same format."
```
akashenfelter: Missed a word: "Can be **used** to complete future datasets of the same format."

repair-missing/script.whizzml
```diff
@@ -0,0 +1,192 @@
...
+        final-ds)
+      dataset-id)))
+
+;; Helper function to get the ID of a field given its name
+(define (id-from-fields fields name)
+  (loop (ids (keys fields) vs (values fields))
+    (cond (empty? ids) false
+          (= name ((head vs) "name" false)) (head ids)
+          (recur (tail ids) (tail vs)))))
+
```
akashenfelter, line 4: This may be a newbie question, but why did you go with a loop here instead of a filter?

repair-missing/script.whizzml
```diff
@@ -0,0 +1,192 @@
+;; This is a hack:  Field IDs are not conserved at batch prediction time, so
+;; we need to make sure that all of the fields have assigned IDs exactly as
+;; batch prediction will assign them (which essentially amounts to putting
+;; generated fields at the end).  To do this, we make a dummy model, do a dummy
+;; batch prediction, then remove the dummy column.
+(define (flatten-dataset dataset-id)
+  (let (fds ((fetch dataset-id) "fields"))
+    (if (some (lambda (fid) (= ((fds fid) "optype") "datetime")) (keys fds))
```
akashenfelter: But the `datetime` optype isn't necessarily present in all datasets with generated data, so those datasets won't get flattened and the batchprediction will still mess up their field ids.  Is there something I'm missing?

repair-missing/readme.md
```diff
@@ -0,0 +1,62 @@
...
+Authentication](http://bigmler.readthedocs.io/en/latest/#bigml-authentication)
+section of docs for details.
+
+Once the setup is complete, go to the test directory and run the shell
+script
+
+```
+bash
+cd test
+./test.sh
```
akashenfelter: Maybe add in a `VERBOSITY` value? But this is wonderful and I am stealing your language for my test readmes.

repair-missing/script.whizzml
```diff
@@ -0,0 +1,192 @@
...
+;; Compose a dataset containing only rows for which the specified
+;; field ID isn't missing
+(define (filter-missing dataset-id field-id)
+  (let (fexpr (str "(not (missing \"" field-id "\"))"))
+    (create-dataset {"origin_dataset" dataset-id "lisp_filter" fexpr})))
+
+;; Create a model with the given dataset for the given field ID if it
+;; is a modelable type, else return the empty string as a dummy
+;; resource ID.
+(define (make-model ds-name train-ds fid fdesc obj-id)
```
akashenfelter, line 38: And another newbie question: why are you passing the field id and field descriptions as two separate variables? To save lookup times?

akashenfelter (changes requested):
One typo and a few questions that my be more about my understanding than your code. Thanks for being my first code review!

repair-missing/readme.md
charleslparker: I stole this almost completely from @mmerce 's Brouta feature selection script!  I also stole most of the test script itself so I'm not sure where or how the verbosity value usually works.  I don't see it in `anomaly-benchmarking` either?  Can you point me to this, or am I just missing it?

repair-missing/script.whizzml
charleslparker: No, I just need both pieces of information in the function.  I don't see a particularly cleaner alternative . . . unless you're suggesting I look up the descriptor from the `train-ds` already sitting there?  In that case, yes, I was just trying to save network call to pull down the dataset resource (you see below that this function is used in a `map`, so the time might be non-trivial depending on the list size).  Are you thinking of something else?

repair-missing/script.whizzml
charleslparker: Not at all a newbie question!  I copy-pasted this from a time before we built in `filter`, which would indeed be better here.  Though in this case (selecting one item), I prefer `some`.  I'll change it.

repair-missing/script.whizzml
charleslparker: There's almost certainly something _I'm_ missing.  I thought datetimes were the only time we were generating fields, but you're saying there are others?  What are they?

repair-missing/metadata.json
charleslparker: Indeed

charleslparker:
A few return questions and a few fixes.  Also, since it looks like @jaor is going to push out the generated fields fix before the next prod release, let's wait to merge this until after the release, and I'll just dump my (apparently broken) hack around that.

repair-missing/script.whizzml
akashenfelter: I'm not sure if it counts as "generated", but my own troubles with the batchprediction name switcharoo are with an extended dataset that doesn't have a `datetime`. For example, take a dataset with fields 000000 through 00000e, and use a flatline expression to add a new field. By default, it will have id 10000f. Use that as the objective field of a logistic regression, and get the output dataset from a batchprediction, and now it's 00000f. 

But yes, jao seems to have fixed this! Wooo!

repair-missing/script.whizzml
akashenfelter: Ah yes, `some` is even better.

repair-missing/script.whizzml
akashenfelter: That's what I was thinking, looking it up from the the `train-ds`. Which, now that you point it out, is a lot of calls. This is something I'm learning to pay attention to when coding, thanks for clarifying!

repair-missing/readme.md
akashenfelter: This messed me up when I learned about running tests. When you run `./test.sh`, you don't get much output. Oftentimes, just an error message if the resources don't clean up in a neat order. With `VERBOSITY=1 ./test.sh`, you see all the lovely log messages, and with `VERBOSITY=2 ./test.sh` you'll see every little thing bigmler is doing. So I ran `./test.sh` and thought my test was bad, but with `VERBOSITY=1 ./test.sh` I could see it was running fine.

And I should really go improve that in `anomaly-benchmarking` too.

akashenfelter:
Sounds great! I'll merge the changes then.

jaor:
re verbosity, that when everything goes well there's not ouput at all is
_by design_: i only want to see output in case of errors (the
unfortunate fact the there's still a bit of output when there are no
errors is, well, unfortunate, and we should check and see if bigmler can
clean up resources without complaining).

so please, keep things quiet when VERBOSITY is 0 :)

akashenfelter:
Yes, absolutely! Maybe it's just a personal quirk (unfamiliarity with running tests?) that I always set VERBOSITY to 1 and like to see the message that everything went ok. There is definite value in having the default be silence when everything went right. No news is good news. :)

repair-missing/script.whizzml
charleslparker: Ah, yes.  Now that you mention this I remember seeing it before.

repair-missing/readme.md
charleslparker: Wow - I was thinking I had to make some changes to make this happen, but I just ran `VERBOSITY=2 ./test.sh` and auto-magically got the extra output!  I'm still not sure exactly how this happened but something about gift horses and mouths applies here.  I added a bit in the readme to let people know about the option.

charleslparker:
> Sounds great! I'll merge the changes then.

Please wait until I give one final go-ahead; once it's tested I'll remove the unncessary functions and let you know when I'm done.  Thanks!

mmerce:
@akashenfelter, we're alike. When I run the tests I expect them to print some output to be sure of what they're doing. This must be due to my controller personality :-). However, @jaor is usually only interested in them if they fail, so he set this VERBOSITY behaviour.

In BigMLer, --verbosity 1 is the default and shows the output and --verbosity 0 is silent. You can even see the requests and response for the http calls if you add --debug at the end. If that's something you could need, we could even add another VERBOSITY code for that.

charleslparker:
Okay, just tested against labs without flattening and everything seems to work.  This is ready to be merged once prod has the batch prediction fix in place.

akashenfelter:
Wintermute is updated, merging.

PR #26, opened 2016-10-17
Added pkg installation instructions to readmes

akashenfelter:
I've knocked around several ways of explaining this, and this seems the least bad. Any improvements to wording are very welcome.

### DIFF
--- a/best-k/readme.md
+++ b/best-k/readme.md
@@ -17,6 +17,9 @@ and three scripts to apply the algorithm to
 which you can install individually afterwards.  When importing any of
 the scripts, `Best K-Means` must be declared as a dependency.
 
+If necessary, please see the top-level readme for more detailed
+installation instructions.
+
 ## Routines
 The library is organized into several functions described next.
 
--- a/best-k/readme.md
+++ b/best-k/readme.md
@@ -17,8 +17,8 @@ and three scripts to apply the algorithm to
 which you can install individually afterwards.  When importing any of
 the scripts, `Best K-Means` must be declared as a dependency.
 
-If necessary, please see the top-level readme for more detailed
-installation instructions.
+If necessary, please see [the top-level readme](../readme.md) for more
+detailed installation instructions.
 
 ## Routines
 The library is organized into several functions described next.
--- a/readme.md
+++ b/readme.md
@@ -55,10 +55,20 @@ By convention, when the artifact is a library, the files are called
   or regression) and turns it into a binary classification problem
   with the two classes "normal" and "anomalous".
 
-## How to install WhizzML packages
+## How to install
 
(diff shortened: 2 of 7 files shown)


best-k/readme.md
```diff
@@ -17,6 +17,9 @@ and three scripts to apply the algorithm to
 which you can install individually afterwards.  When importing any of
 the scripts, `Best K-Means` must be declared as a dependency.
 
+If necessary, please see the top-level readme for more detailed
+installation instructions.
```
jaor: you can make those references to "the top-level readme" a link with 

``` markdown
[the top-level readme](../readme.md)
```

readme.md
```diff
@@ -55,6 +55,35 @@ By convention, when the artifact is a library, the files are called
...
+depend on. They can be installed in either of the following ways:
+
+### Using bigmler
+
+If you have bigmler installed in your system, just checkout the
+repository 'whizzml/examples' and, at its top level, issue the command:
+
+        make compile PKG=example-name
+
+replacing "example-name" with the actual example name. That will
```
jaor: silly nit, but i'd use here

``` markdown
`example-name`
```

instead of quotes.

readme.md
```diff
@@ -55,6 +55,35 @@ By convention, when the artifact is a library, the files are called
   or regression) and turns it into a binary classification problem
   with the two classes "normal" and "anomalous".
 
+## How to install WhizzML packages
```
jaor: i would write some kind of introduction here saying that there are three kinds of installable whizzml artifacts in this repo, identified by the field "kind" in their metadata: libraries, scripts and packages.  The latter are compounds of libraries and scripts, possibly with interdependencies, meant to be installed together.

readme.md
```diff
@@ -55,6 +55,35 @@ By convention, when the artifact is a library, the files are called
...
+  their folders. (For example, https://github.com/whizzml/examples/tree/master/clean-dataset/clean-data)
+  
+- Install each of the scripts seperately, using the
+  urls to each of their folders.  
+
+- If a script requires a library, you will get the error message
+  'Library ../my-library not loaded.' Load the library by clicking in
+  the textbox above the error message and typing the first few letters
+  of the library's name. Select the library, then create the script as
+  usual.
```
jaor: this is correct, but maybe we could nag the frontend guys so that the error is bit nicer (i remember that when you select the library, the message doesn't dissappear, no?)

jaor (changes requested):

akashenfelter:
Nits picked, silly ones and all.  I also added some instructions for installing plain old scripts and libraries.

akashenfelter:
Agreed, that error message is not terribly helpful. And you are correct, it
doesn't disappear after selecting the library. Who should I mention this to?

Allison

PR #27, opened 2016-10-24
Better delete in SMACdown

charleslparker:


charleslparker:
@jaor - Hold on, I'm adding the other delete thing

smacdown-ensemble/script.whizzml
```diff
@@ -61,6 +61,12 @@
   (try (delete id)
        (catch e (log-info (str "Error deleting resource " id " ignored")))))
 
+(define (delete-batchprediction-datasets resources)
+  (let (ids (filter (lambda (id) (!= id ""))
+                    (map (lambda (r) ((fetch r) "output_dataset_resource" ""))
+                         resources)))
+    (map safe-delete ids)))
```
jaor: since whizzml is not specially quick, maybe a single `map` would be
enough? something like:

```
(for (r resources)
  (let (id ((fetch r) "output_dataset_resource" false))
    (when id (safe-delete id))))
```

(i also think it's a bit easier to read)

jaor (commented):
just a nit....

charleslparker:
Yes, I like yours better.  Should be okay now.

jaor:
okay, looks good to me! @osroca would you mind trying the new version before we merge to make sure it behaves as you expected? thanks!

charleslparker:
Also, note, @osroca, that it looks like the version you were using has some modifications that aren't  in this repo.  I'm not sure where those came from or if they matter.

osroca:
what are the differences? we updated the description of the script and also the description of the input `metric` to let the users know which can be used. Can you update them as well, please? you can find these here https://bigml.io/andromeda/script/580b79b30144047ae7000a64?username=osroca;api_key=8ea8aba1b2c491dd5117f3379a4a6803f411da15

charleslparker:
> what are the differences? 

@osroca - The ones you mention are the only ones I could see, but it means someone has been modifying the script without a PR to this repo so I'm not totally sure if everything else is the same.

osroca:
@charleslparker the descriptions I had provided in my PR on your repo are the ones we need. The general script description supports marcdown and thus we want to list the possible metrics using an unordered list (the `\n` are needed in order to get the correct format of the output). Regarding the description of the input `metric`, it is presented in a tooltip that truncates the content, the one I provided fits well on the limits. Then, you can of course change the English, but please keep the format and the length.

Looking at your proposal, it look like for Regression problems SMACdown only optimizes `r_squared`, is that correct?

examples of how we display it in the UI.
image
image

your `metric` description overpass the tooltip limit
image

charleslparker:
> @charleslparker the descriptions I had provided in my PR on your repo are the ones we need. The general script description supports marcdown and thus we want to list the possible metrics using an unordered list (the \n are needed in order to get the correct format of the output). Regarding the description of the input metric, it is presented in a tooltip that truncates the content, the one I provided fits well on the limits. Then, you can of course change the English, but please keep the format and the length.

Ah, I see.  Yes, I'll change to to look similar to yours.

> Looking at your proposal, it look like for Regression problems SMACdown only optimizes r_squared, is that correct?

It is, unless there was a PR that I missed.  It requires a metric where _higher is better_, and of course mean absolute/squared error are the opposite.

osroca:
> > Looking at your proposal, it look like for Regression problems SMACdown only optimizes r_squared, is that correct?
> > It is, unless there was a PR that I missed. It requires a metric where higher is better, and of course mean absolute/squared error are the opposite.

then, my descriptions are incorrect ... please fix them.

jaor:
@charleslparker, sorry, i've lost track of this one: what's the status?

charleslparker:
It should be all set.

smacdown-ensemble/script.whizzml
```diff
@@ -61,6 +61,12 @@
   (try (delete id)
        (catch e (log-info (str "Error deleting resource " id " ignored")))))
 
+(define (delete-batchprediction-datasets resources)
+  (for (r resources)
+       (when (= (resource-type r) "batchprediction")
```
jaor: sorry i didn't see this before, but the indentation is wrong here: should be:

```
(for (r resources)
  (when
```

whizzmler:
added a silly nit...

charleslparker:
Serves me right for not using whizzml-mode!

PR #28, opened 2016-11-07
anomaly-benchmarking: added generate-datasets

akashenfelter:
Here is the working implementation of the ODD paper. I made one minor change from the paper -- this script creates ten replicate datasets instead of 40. Special thanks to @charleslparker, for his resource deletion in the clean-data example.

Of note, I was running into some problems with queueing up too many datasets at once. This was "solved" by using `create-and-wait-dataset` even though I don't need to rely on that resource to continue.  It would be nice to not have to worry about parallelization when writing a whizzml script.

### DIFF
--- a/anomaly-benchmarking/generate-datasets/metadata.json
+++ b/anomaly-benchmarking/generate-datasets/metadata.json
@@ -1,6 +1,6 @@
 {
-  "name": "Generate datasets",
-  "description": "",
+  "name": "Generate benchmarking datasets",
+  "description": "Creates a suite of benchmark datasets for anomaly detection algorithms",
   "kind": "script",
   "source_code": "script.whizzml",
   "imports": ["../make-binary"],
--- a/anomaly-benchmarking/generate-datasets/metadata.json
+++ b/anomaly-benchmarking/generate-datasets/metadata.json
@@ -1,0 +1,27 @@
+{
+  "name": "Generate datasets",
+  "description": "",
+  "kind": "script",
+  "source_code": "script.whizzml",
+  "imports": ["../make-binary"],
+  "inputs": [
+    {
+      "name": "dataset-list",
+      "type": "list",
+      "description": "The resource ids of the 'mother' datasets"
+    },
+    {
+      "name": "delete-resources",
+      "type": "boolean",
+      "default": true,
+      "description": "Whether to delete all intermediate resources"
+    }
+  ],
+  "outputs": [
+    {
+      "name": "generated-datasets",
+      "type": "list",
+      "description": "generated children datasets"
+    }
+  ]
+}
--- a/anomaly-benchmarking/generate-datasets/readme.md
+++ b/anomaly-benchmarking/generate-datasets/readme.md
@@ -1,0 +1,50 @@
+# Generating Benchmarking Datasets for Anomaly Detection
+
+This script takes a list of datasets and transforms them into a suite
+of ground-truthed benchmark datasets for anomaly detection as laid out
+in the ODD paper
+http://www.outlier-analytics.org/odd13kdd/papers/emmott,das,dietterich,fern,wong.pdf. It depends on the library "make-binary". (See [readme](../make-binary/readme.md))
+
+These output datasets vary along the dimensions point difficulty, relative
+frequency of anomalies, and clusteredness.
+
+For each dataset id in the list, this script: 
+
+- Transforms the dataset into a binary classification using `make-binary`.
+
+- Seperates out the "normal" rows.
+
+- Point difficulty: Models the dataset with a logistic regression, and
+  uses a batchprediciton to labels each row with the probability of it
+  being labeled "normal". The rows labeled "anomalous" are sorted into
+  one of four datasets based on this probability. ("Easy" (0, 0.1666),
+  "medium" [0.1666, 0.3333), "hard" [0.3333, 0.5), "very hard" [0.5,
+  1))
+
+- Clusteredness: For each of the point difficulty datasets, an anomaly
+  detector (isolation forest) and batchprediction are used to label
+  each row with an anomaly score. This is a rough gauge for
+  clusteredness, and new datasets are created from the top and bottom
+  20% anomaly scores. The actual clusteredness is calculated by
+  finding the ratio of the variance of the "normal" rows to the
+  variance of the new dataset (which has N rows). Actual clusteredness
+  is sorted into one of six categories: high scatter (0, 0.25), medium
+  scatter [0.25, 0.5), low scatter [0.5, 1), low clusteredness [1,2),
+  medium clusteredness [2, 4), and high clusteredness [4, infinity).
+
+- Frequency: As long as the dataset has at least 10 rows, a new
+  dataset (which has K rows) is created through a sample rate set to
+  create the goal frequency. Goal frequencies are 0.001, 0.005, 0.01,
+  0.05, and 0.1.
+
+- If enough rows are availible, ten replicate datasets are created at
+  each combination of difficulty, clusteredness, and frequency. A
+  maxiumum of N/K replicates are created.
+
+- These datasets are merged back with the "normal" rows.
+
+- The final output is a list of maps with keys "difficulty",
+  "frequency", "variance", and "resource" (the resource id of the
+  final merged dataset).
+
+If `delete-resources` is true, it deletes all intermediate resources.
--- a/anomaly-benchmarking/generate-datasets/readme.md
+++ b/anomaly-benchmarking/generate-datasets/readme.md
@@ -10,14 +10,15 @@ frequency of anomalies, and clusteredness.
 
 For each dataset id in the list, this script: 
 
-- Transforms the dataset into a binary classification using `make-binary`.
+- Transforms the dataset into a binary classification using the
+  library function `make-binary`.
 
 - Seperates out the "normal" rows.
 
 - Point difficulty: Models the dataset with a logistic regression, and
-  uses a batchprediciton to labels each row with the probability of it
+  uses a batchprediction to labels each row with the probability of it
   being labeled "normal". The rows labeled "anomalous" are sorted into
-  one of four datasets based on this probability. ("Easy" (0, 0.1666),
+  one of four datasets based on this probability. ("easy" (0, 0.1666),
   "medium" [0.1666, 0.3333), "hard" [0.3333, 0.5), "very hard" [0.5,
   1))
 
--- a/anomaly-benchmarking/generate-datasets/script.whizzml
+++ b/anomaly-benchmarking/generate-datasets/script.whizzml
@@ -145,7 +145,7 @@
 
 (define (generate-map ds-id normal-ds)
   (let (ds (fetch ds-id)
-        combined (create-and-wait-dataset {"origin_datasets" [ds-id normal-ds]}) 
+        combined (create-dataset {"origin_datasets" [ds-id normal-ds]}) 
         my-tags (ds "tags"))
     (make-map ["difficulty" "frequency" "variation"  "resource"] 
               (append my-tags combined))))
--- a/anomaly-benchmarking/generate-datasets/script.whizzml
+++ b/anomaly-benchmarking/generate-datasets/script.whizzml
@@ -2,43 +2,48 @@
 
 (define (all-preferred ds-id)
   (let (ds (fetch ds-id)
-        field-list (keys (ds "fields"))
-        not-preferred (filter (lambda (x) (not (get-in ds ["fields" x "preferred"]))) field-list))
-     (update ds-id {"fields" (make-map not-preferred (repeat (count not-preferred) {"preferred" true}))})))
+        field-list (keys (ds "fields")))
+    (update ds-id 
+            {"fields" (make-map field-list (repeat (count field-list) 
+                                                   {"preferred" true}))})))
 
 ;; Given a dataset and a field name, returns the field id with that name
 
 (define (get-id ds name)
-  (let (field (ds "fields"))
-    (head (filter (lambda (x) (= (field [x "name"]) name)) 
-                (keys field)))))
+  (let (fields (ds "fields"))
+    (try
+      (head (filter (lambda (x) (= (fields [x "name"]) name)) 
+                    (keys fields)))
+      (catch e
+        (log-warn "Could not find field name " name " : " e)))))
 
 ;; Given a binary dataset, finds the point difficulty of each row and
-;; returns an dataset of only anomalous points and a new field
+;; returns a dataset of only anomalous points and a new field
 ;; labeling each row by one of four difficulty levels.
 
-(define (point-difficulty ds-id delete-resources)
+(define (point-difficulty ds-id new-obj delete-resources)
   (let (ds (all-preferred ds-id)
-        lr-id (create-and-wait-logisticregression {"dataset" ds})
-        lr-fields ((fetch lr-id) "input_fields")
-        bp-id (create-and-wait-batchprediction {"logisticregression" lr-id 
-                                       "dataset" ds 
-                                       "probabilities" true 
-                                       "output_dataset" true 
-                                       "output_fields" lr-fields })
+        lr-id (create-and-wait-logisticregression ds)
+        bp-id (wait (create-batchprediction lr-id ds {"probabilities" true 
+                                                      "output_dataset" true 
+                                                      "all_fields" true }))
         prob-ds-id ((fetch bp-id) "output_dataset_resource")
-        prob-ds (fetch prob-ds-id)
+        prob-ds (fetch (wait prob-ds-id))
         norm-field (get-id prob-ds "normal probability")
         ano-field (get-id prob-ds "anomalous probability")
-        obj-field (get-id prob-ds "Is Anomaly?")
-        difficulty-sorter (lambda (diff-id) (flatline "(segment-label {{diff-id}} \"easy\" 0.133333 \"medium\" 0.333333 \"hard\" 0.5 \"very hard\")"))
-        diff-ds (create-and-wait-dataset {"origin_dataset" prob-ds-id
-                  "lisp_filter" (flatline "(= (field {{obj-field}}) \"anomalous\")")
-                  "new_fields" [{"field" (difficulty-sorter norm-field)
-                                 "name" "sorted"}]
-                  "all_but" [norm-field ano-field]}))
+        diff-sort (flatline "(segment-label {{norm-field}}" 
+                            "               \"easy\" 0.133333"
+                            "               \"medium\" 0.333333"
+                            "               \"hard\" 0.5"
+                            "               \"very hard\")")
+        my-filter (flatline "(= (field {{new-obj}}) \"anomalous\")")
+        diff-ds (wait (create-dataset prob-ds-id 
+                                      {"lisp_filter" my-filter
+                                       "new_fields" [{"field" diff-sort
+                                                      "name" "sorted"}]
+                                       "all_but" [norm-field ano-field]})))
     (when delete-resources
-          (map safe-delete [lr-id bp-id prob-ds-id]))
+      (map safe-delete [lr-id bp-id prob-ds-id]))
     diff-ds))
 
 ;; Takes a dataset labeled by point-difficulty, and splits the dataset
@@ -49,17 +54,18 @@
   (let (ds (fetch ds-id)
         split-field (get-id ds "sorted")
         split-list (ds ["fields" split-field "summary" "categories"]) 
-        split-map (make-map (map (lambda (x) (head x)) split-list) 
-                            (map (lambda (x) (last x)) split-list))
+        split-map (make-map (map head split-list)(map last split-list))
         split-it (lambda (level)
-                 (if (contains? split-map level) 
-                   (create-and-wait-dataset {"origin_dataset" ds-id
-                     "lisp_filter" (flatline "(= (field {{split-field}}) {{level}})")
-                     "excluded_fields" [split-field]
-                     "name" (str level " difficulty")
-                     "tags" [level]})
-                   false)))
-    (for (x ["easy" "medium" "hard" "very hard"]) (split-it x))))
+                   (let (my-filter (flatline "(= (f {{split-field}})" 
+                                             "   {{level}})"))
+                     (if (contains? split-map level) 
+                       (wait (create-dataset {"origin_dataset" ds-id
+                                              "lisp_filter" my-filter
+                                              "excluded_fields" [split-field]
+                                              "name" (str level " diff")
+                                              "tags" [level]}))
+                       false))))
+    (map split-it ["easy" "medium" "hard" "very hard"])))
 
 ;; Takes a dataset and creates two datasets, one from the lowest
 ;; anomaly scores and one from the highest. Returns a list of the
@@ -70,24 +76,26 @@
         ds-name (ds "name")
         ad-id (create-anomaly {"dataset" ds-id
                                "id_fields" [obj-field]})
-        bas-id (create-batchanomalyscore {"anomaly" (wait ad-id)
+        bas-id (create-batchanomalyscore {"anomaly" ad-id
                                           "dataset" ds-id
                                           "all_fields" true
                                           "output_dataset" true})
         score-ds ((fetch (wait bas-id)) "output_dataset_resource")
-        id (get-id (fetch score-ds) "score")
+        id (get-id (fetch (wait score-ds)) "score")
         score-low (flatline "(within-percentiles? {{id}} 0 0.2)")
         score-high (flatline "(within-percentiles? {{id}} 0.8 1)")
         low-ds (create-and-wait-dataset {"origin_dataset" score-ds
-                                "lisp_filter" score-low
-                                "name" (str ds-name " - high clusteredness")
-                                "excluded_fields" [id]})
+                                         "lisp_filter" score-low
+                                         "name" (str ds-name 
+                                                     " - high clusteredness")
+                                         "excluded_fields" [id]})
         high-ds (create-and-wait-dataset {"origin_dataset" score-ds
-                                 "lisp_filter" score-high
-                                 "name" (str ds-name " - low clusteredness")
-                                 "excluded_fields" [id]}))
+                                          "lisp_filter" score-high
+                                          "name" (str ds-name 
+                                                      " - low clusteredness")
+                                          "excluded_fields" [id]}))
     (when delete-resources
-          (map safe-delete [ad-id bas-id score-ds]))
+      (map safe-delete [ad-id bas-id score-ds]))
     [low-ds high-ds]))
 
 ;; Takes two datasets, one normal and one anomalous. Finds the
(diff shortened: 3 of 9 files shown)


jaor:
@akashenfelter when you say you were running into problems with queued datasets, was it executing in the server? if so, it'd be a bug in our server side and i'd like to investigate (if it's in the client side, that's a limitation of the client repl and cannot be solve, but the script should _not_ use create-and-wait, since it's really meant to run on the server).

jaor:
(which means that, no matter what, we should remove those create-and-wait from this script)

akashenfelter:
I got the error `Error computing primitive operation 'create'. Error handling resource (You don't have enough credits available)`. Talking with @petersen-poul, he says that happens if you exceed the parallelization. I'm not sure it's a bug, as such. I can see why we wouldn't want an arbitrarily long queue.

akashenfelter:
But I would also welcome being able to create as many datasets as I want, and let the server figure out how to make them. :)

whizzmler:
the server should already being figuring out how to parallelize your
tasks without problems.  the "you don't have enough credits available"
error might be related to the size of your jobs, not the level of
parallelism.  what are your subscription's parameters? (you didn't tell
me if you got the errors through the REPL or in an execution, but i'm
guessing the latter).  if your subscription parameters are fine, then
there's a bug in our server.

akashenfelter:
With further investigation, the queueing problem was when trying to run the script in the java repl, not as an execution. Everything is fine when run directly as packaged. I'm removing the `-and-wait-`.

anomaly-benchmarking/generate-datasets/metadata.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "Generate datasets",
```
jaor: perhaps "Generate benchmarking datasets"?

anomaly-benchmarking/generate-datasets/metadata.json
```diff
@@ -0,0 +1,27 @@
...
+  "description": "",
```
jaor: and a one-line summary description?

anomaly-benchmarking/generate-datasets/readme.md
```diff
@@ -0,0 +1,50 @@
...
+frequency of anomalies, and clusteredness.
+
+For each dataset id in the list, this script: 
+
+- Transforms the dataset into a binary classification using `make-binary`.
+
+- Seperates out the "normal" rows.
+
+- Point difficulty: Models the dataset with a logistic regression, and
+  uses a batchprediciton to labels each row with the probability of it
```
jaor: typo: batchprediction

anomaly-benchmarking/generate-datasets/readme.md
```diff
@@ -0,0 +1,50 @@
...
+of ground-truthed benchmark datasets for anomaly detection as laid out
+in the ODD paper
+http://www.outlier-analytics.org/odd13kdd/papers/emmott,das,dietterich,fern,wong.pdf. It depends on the library "make-binary". (See [readme](../make-binary/readme.md))
+
+These output datasets vary along the dimensions point difficulty, relative
+frequency of anomalies, and clusteredness.
+
+For each dataset id in the list, this script: 
+
+- Transforms the dataset into a binary classification using `make-binary`.
```
jaor: perhaps mention that's a library function

anomaly-benchmarking/generate-datasets/readme.md
```diff
@@ -0,0 +1,50 @@
...
+
+- Seperates out the "normal" rows.
+
+- Point difficulty: Models the dataset with a logistic regression, and
+  uses a batchprediciton to labels each row with the probability of it
+  being labeled "normal". The rows labeled "anomalous" are sorted into
+  one of four datasets based on this probability. ("Easy" (0, 0.1666),
```
jaor: if the actual label is "easy" i don't think we should capitalize here...

anomaly-benchmarking/generate-datasets/script.whizzml
```diff
@@ -0,0 +1,205 @@
+;; Takes a dataset, returns the same dataset with all fields marked preferred
+
+(define (all-preferred ds-id)
+  (let (ds (fetch ds-id)
+        field-list (keys (ds "fields"))
+        not-preferred (filter (lambda (x) (not (get-in ds ["fields" x "preferred"]))) field-list))
+     (update ds-id {"fields" (make-map not-preferred (repeat (count not-preferred) {"preferred" true}))})))
```
jaor: i wouldn't filter non-preferred, just mark all of them as preferred unconditionally:

```
(update ds-id
        {"fields" (make-map (keys fields)
                            (repeat (count fields)
                                    {"preferred" true}))})
```

which is, imo, shorter and easier to understand at first glance.  if you insist on filtering, you should probably guard yourself against an empty `not-preferred` list (and break-down that update to keep the 80 cols limit).

anomaly-benchmarking/generate-datasets/script.whizzml
```diff
@@ -0,0 +1,205 @@
...
+
+;; Given a dataset and a field name, returns the field id with that name
+
+(define (get-id ds name)
+  (let (field (ds "fields"))
+    (head (filter (lambda (x) (= (field [x "name"]) name)) 
+                (keys field)))))
```
jaor: indentation is off, and perhaps a bit confusing you name the fields map `field` instead of `fields`.

jaor: also, this is going to throw an ugly error if the field is not found... perhaps you'll want to check and `raise` a nicer message...

anomaly-benchmarking/generate-datasets/script.whizzml
```diff
@@ -0,0 +1,205 @@
...
+
+;; Given a binary dataset, finds the point difficulty of each row and
+;; returns an dataset of only anomalous points and a new field
```
jaor: typo: an -> a

anomaly-benchmarking/generate-datasets/script.whizzml
```diff
@@ -0,0 +1,205 @@
...
+;; labeling each row by one of four difficulty levels.
+
+(define (point-difficulty ds-id delete-resources)
+  (let (ds (all-preferred ds-id)
+        lr-id (create-and-wait-logisticregression {"dataset" ds})
```
jaor: `(create-and-wait-logistic-regression ds)` works just the same and it's a bit shorter

anomaly-benchmarking/generate-datasets/script.whizzml
```diff
@@ -0,0 +1,205 @@
...
+        lr-fields ((fetch lr-id) "input_fields")
+        bp-id (create-and-wait-batchprediction {"logisticregression" lr-id 
+                                       "dataset" ds 
+                                       "probabilities" true 
+                                       "output_dataset" true 
+                                       "output_fields" lr-fields })
```
jaor: you can just use `"all_fields" true`, and then you don't have to wait for the lr and get its input fields, so the whole thing reads:

```
lr-id (create-logisticregression ds)
bp-id (wait (create-batchprediction lr-id ds {"probabilities" true
                                              "output_dataset" true
                                              "all_fields" true}))
```

there i'm just using wait because i like it that way but
create-and-wait-batchprediction would be perfectly fine, and i'm
passing the raw ids.

anomaly-benchmarking/generate-datasets/script.whizzml
```diff
@@ -0,0 +1,205 @@
...
+        prob-ds-id ((fetch bp-id) "output_dataset_resource")
+        prob-ds (fetch prob-ds-id)
```
jaor: you'd better wait for this one, the dataset is not guaranteed to be finished when the batch finishes (and fetch, unlike the create functions, won't wait): `(fetch (wait prob-ds-id))`

anomaly-benchmarking/generate-datasets/script.whizzml
```diff
@@ -0,0 +1,205 @@
...
+        norm-field (get-id prob-ds "normal probability")
+        ano-field (get-id prob-ds "anomalous probability")
+        obj-field (get-id prob-ds "Is Anomaly?")
```
jaor: this "Is Anomaly?" looks like an excellent candidate for a constant defined in the imported library...

anomaly-benchmarking/generate-datasets/script.whizzml
```diff
@@ -0,0 +1,205 @@
...
+        difficulty-sorter (lambda (diff-id) (flatline "(segment-label {{diff-id}} \"easy\" 0.133333 \"medium\" 0.333333 \"hard\" 0.5 \"very hard\")"))
```
jaor: line too long.  note that flatline takes multiple strings, so you can nicely format the quoted expression....e.g.:

```
difficulty-sorter (lambda (diff-id)
                    (flatline "(segment-label {{diff-id}}"
                              "               \"easy\" 0.133333"
                              "               \"medium\" 0.33333"
                              "               \"hard\" 0.5"
                              "               \"very hard\")"))
```

again, maybe the category names would be a candidate for constants
defined somwhere, and you could use there things like `{{C-EASY}}`...

anomaly-benchmarking/generate-datasets/script.whizzml
```diff
@@ -0,0 +1,205 @@
...
+        diff-ds (create-and-wait-dataset {"origin_dataset" prob-ds-id
+                  "lisp_filter" (flatline "(= (field {{obj-field}}) \"anomalous\")")
```
jaor: indentation looks off here too.  again, you could just `(create-and-wait-dataset prob-ds-id ...`

anomaly-benchmarking/generate-datasets/script.whizzml
```diff
@@ -0,0 +1,205 @@
...
+                  "new_fields" [{"field" (difficulty-sorter norm-field)
+                                 "name" "sorted"}]
+                  "all_but" [norm-field ano-field]}))
+    (when delete-resources
+          (map safe-delete [lr-id bp-id prob-ds-id]))
```
jaor: wrong indentation here too: `when` has a body after the check:

```
(when foo
  bar
  baz)
```

anomaly-benchmarking/generate-datasets/script.whizzml
```diff
@@ -0,0 +1,205 @@
...
+;; Takes a dataset labeled by point-difficulty, and splits the dataset
+;; into one of four new datasets based on their label. Returns the
+;; list of these new datasets.
+
+(define (split-difficulty ds-id)
+  (let (ds (fetch ds-id)
+        split-field (get-id ds "sorted")
+        split-list (ds ["fields" split-field "summary" "categories"]) 
+        split-map (make-map (map (lambda (x) (head x)) split-list) 
+                            (map (lambda (x) (last x)) split-list))
```
jaor: never wrap a function of one argument with a lambda, it's not necessary:

```
(make-map (map head split-list) (map last split-list))
```

anomaly-benchmarking/generate-datasets/script.whizzml
```diff
@@ -0,0 +1,205 @@
...
+        split-it (lambda (level)
+                 (if (contains? split-map level)
```
jaor: indentation

anomaly-benchmarking/generate-datasets/script.whizzml
```diff
@@ -0,0 +1,205 @@
...
+(define (split-difficulty ds-id)
+  (let (ds (fetch ds-id)
+        split-field (get-id ds "sorted")
+        split-list (ds ["fields" split-field "summary" "categories"]) 
+        split-map (make-map (map (lambda (x) (head x)) split-list) 
+                            (map (lambda (x) (last x)) split-list))
+        split-it (lambda (level)
+                 (if (contains? split-map level) 
+                   (create-and-wait-dataset {"origin_dataset" ds-id
+                     "lisp_filter" (flatline "(= (field {{split-field}}) {{level}})")
```
jaor: indentation...

```
(create-and-wait-dataset {"foo" 3
                          "bar" 4
                          "baz" 5})
```

anomaly-benchmarking/generate-datasets/script.whizzml
```diff
@@ -0,0 +1,205 @@
...
+                     "excluded_fields" [split-field]
+                     "name" (str level " difficulty")
+                     "tags" [level]})
+                   false)))
+    (for (x ["easy" "medium" "hard" "very hard"]) (split-it x))))
```
jaor: shorter alternative (but your's is fine if you like it better):

```
(map split-it ["easy" "medium" "hard" "very hard"])
```

(if that list of difficulties appears elsewhere, i'd have a variable for it)

anomaly-benchmarking/generate-datasets/script.whizzml
```diff
@@ -0,0 +1,205 @@
...
+;; Takes a dataset and creates two datasets, one from the lowest
+;; anomaly scores and one from the highest. Returns a list of the
+;; dataset ids.
+
+(define (semantic-variation ds-id obj-field delete-resources)
+  (let (ds (fetch ds-id)
+        ds-name (ds "name")
+        ad-id (create-anomaly {"dataset" ds-id
+                               "id_fields" [obj-field]})
+        bas-id (create-batchanomalyscore {"anomaly" (wait ad-id)
```
jaor: don't need the explicit `wait`, create will automatically wait for you.

anomaly-benchmarking/generate-datasets/script.whizzml
```diff
@@ -0,0 +1,205 @@
...
+                                          "dataset" ds-id
+                                          "all_fields" true
+                                          "output_dataset" true})
+        score-ds ((fetch (wait bas-id)) "output_dataset_resource")
+        id (get-id (fetch score-ds) "score")
```
jaor: `score-ds` needs to be waited for too: `id (get-id (fetch (wait score-ds))` ....

anomaly-benchmarking/generate-datasets/script.whizzml
```diff
@@ -0,0 +1,205 @@
...
+        score-low (flatline "(within-percentiles? {{id}} 0 0.2)")
+        score-high (flatline "(within-percentiles? {{id}} 0.8 1)")
+        low-ds (create-and-wait-dataset {"origin_dataset" score-ds
+                                "lisp_filter" score-low
```
jaor: indentation

anomaly-benchmarking/generate-datasets/script.whizzml
```diff
@@ -0,0 +1,205 @@
...
+                                "name" (str ds-name " - high clusteredness")
+                                "excluded_fields" [id]})
+        high-ds (create-and-wait-dataset {"origin_dataset" score-ds
+                                 "lisp_filter" score-high
```
jaor: indentation

anomaly-benchmarking/generate-datasets/script.whizzml
```diff
@@ -0,0 +1,205 @@
...
+                                 "name" (str ds-name " - low clusteredness")
+                                 "excluded_fields" [id]}))
+    (when delete-resources
+          (map safe-delete [ad-id bas-id score-ds]))
```
jaor: indentation

anomaly-benchmarking/generate-datasets/script.whizzml
```diff
@@ -0,0 +1,205 @@
...
+;; to anomalous variance, sorts the ratio into one of six categories,
+;; and tags the anomalous dataset with that category. The variance is
+;; found by abusing k-means; it creates just one cluster which contains
+;; all the data.
+
+(define (variance-ratios ano-ds-id norm-var norm-scales delete-resources)
+  (let (ano-ds (fetch ano-ds-id)
+        my-tags (ano-ds "tags")
+        ano-cl-id (create-and-wait-cluster {"dataset" ano-ds-id 
+                                   "k" 1
```
jaor: indentation

anomaly-benchmarking/generate-datasets/script.whizzml
```diff
@@ -0,0 +1,205 @@
...
+                                    "standard_deviation"])
+        my-ratio (pow (/ norm-var ano-var) 2)
+        new-tag (cond (< my-ratio 0.25) "high scatter"
+                      (< my-ratio 0.5) "medium scatter"
+                      (< my-ratio 1) "low scatter"
+                      (< my-ratio 2) "low clusteredness"
+                      (< my-ratio 4) "medium clusteredness"
+                      "high clusteredness"))
+    (when delete-resources
+          (map safe-delete [ano-cl-id]))
```
jaor: indentation

anomaly-benchmarking/generate-datasets/script.whizzml
```diff
@@ -0,0 +1,205 @@
...
+(define (relative-frequency ds-id normal-rows)
+  (let (ds (fetch ds-id)
+        availible-rows (ds "rows")
+        ds-name (ds "name")
+        my-tags (ds "tags")
+        max-freq (/ availible-rows (+ availible-rows normal-rows)))
+    (for (freq [0.001 0.005 0.01 0.05 0.1])
+      (let (k (/ (* freq normal-rows) (- 1 freq)))
+        (when (> max-freq freq)
+              (repeatedly (min 10 (floor (/ availible-rows k)))
```
jaor: indentation

anomaly-benchmarking/generate-datasets/script.whizzml
```diff
@@ -0,0 +1,205 @@
...
+          (map safe-delete [ano-cl-id]))
+    (update ano-ds-id {"tags" (append my-tags new-tag)})))
+
+;; Takes a dataset. Determines which relative frequencies are
+;; possible. Generates the max possible datasets (limit 40) at those
+;; frequencies. Returns a list of datasets.
+
+(define (relative-frequency ds-id normal-rows)
+  (let (ds (fetch ds-id)
+        availible-rows (ds "rows")
```
jaor: typo: available

anomaly-benchmarking/generate-datasets/script.whizzml
```diff
@@ -0,0 +1,205 @@
...
+        ds-name (ds "name")
+        my-tags (ds "tags")
+        max-freq (/ availible-rows (+ availible-rows normal-rows)))
+    (for (freq [0.001 0.005 0.01 0.05 0.1])
```
jaor, line 140: because of the `when` below, the list evaluated in this `for` may contain `false`s as well as dataset ids: i guess you are taking that into account? an alternative would be to `(filter identity ...)` the list here.

anomaly-benchmarking/generate-datasets/script.whizzml
```diff
@@ -0,0 +1,205 @@
...
+;; clusteredness. Returns a list of maps for all the datasets created,
+;; specifying the resource id, difficulty, semantic variation, and
+;; frequency.
+
+(define (generate ds-id delete-resources)
+  (let (old-obj ((fetch ds-id) ["objective_field" "id"])
+        binary (make-binary ds-id old-obj delete-resources)
+        new-obj ((fetch binary) ["objective_field" "id"])
+        normal-ds (create-and-wait-dataset {"origin_dataset" binary
+                    "lisp_filter"
```
jaor: indentation (it's a chore to check via github, but these indentation problems are not because you are using tabs, right? we hates tabsseeess :))

anomaly-benchmarking/generate-datasets/script.whizzml
```diff
@@ -0,0 +1,205 @@
...
+                           "distance" 
+                           "standard_deviation"])
+        norm-scales (norm-cl "scales")
+        diff-ds (point-difficulty binary delete-resources)
+        difficulty-list (filter (lambda (x) x) (split-difficulty diff-ds))
+        variation-list (flatten [] 
+                                (for (x difficulty-list)
+                                  (if x (semantic-variation x 
+                                                            new-obj 
+                                                        delete-resources))))
```
jaor, line 195: indentation is weird.  if this is a one-legged `if`, use `when` instead.

anomaly-benchmarking/generate-datasets/script.whizzml
```diff
@@ -0,0 +1,205 @@
...
+                                                            new-obj 
+                                                        delete-resources)))) 
+        frequency-list (flatten []
+                                (for (x variation-list)
+                                  (if x (relative-frequency x normal-rows))))
+        labeled-by-variation (for (x frequency-list)
+                               (if x (variance-ratios x 
+                                                      norm-var 
+                                                      norm-scales 
+                                                      delete-resources)))
```
jaor: `when` instead of `if`

anomaly-benchmarking/generate-datasets/script.whizzml
```diff
@@ -0,0 +1,205 @@
...
+        frequency-list (flatten []
+                                (for (x variation-list)
+                                  (if x (relative-frequency x normal-rows))))
+        labeled-by-variation (for (x frequency-list)
+                               (if x (variance-ratios x 
+                                                      norm-var 
+                                                      norm-scales 
+                                                      delete-resources))) 
+        final-list (for (x labeled-by-variation)
+                     (if x (generate-map x normal-ds))))
```
jaor, line 207: `when`

anomaly-benchmarking/generate-datasets/script.whizzml
```diff
@@ -0,0 +1,205 @@
...
+    (when delete-resources
+          (map safe-delete (flatten [] [binary
```
jaor: indentation

anomaly-benchmarking/make-binary/library.whizzml
```diff
@@ -10,20 +10,26 @@
...
+                              "new_fields" [{"field" (two-class-filter 
+                                                      field-id 
+                                                      anomalous)
+                                             "name" "Is Anomaly?"}]
+                              "all_but" [field-id]
+                              "name" (str my-name " child")})))
+
+;; Delete resources ignoring errors
+(define (safe-delete id)
+  (if id (delete id)))
```
jaor: `when`

anomaly-benchmarking/make-binary/library.whizzml
```diff
@@ -10,20 +10,26 @@
...
+;;       (catch e (log-info (str "Error deleting resource " id " ignored")))))
```
jaor: leftover?

anomaly-benchmarking/make-binary/library.whizzml
```diff
@@ -44,11 +50,13 @@
...
 ;; with the remaining 20%. Returns the evaluation id.
 
-(define (evaluate-model-on-dataset ds-id)
+(define (evaluate-model-on-dataset ds-id delete-resources)
   (let (training-id (sample-dataset ds-id 0.8 false)
         test-id (sample-dataset ds-id 0.8 true)
         model-id (create-model {"dataset" training-id "randomize" true})
         ev-id (create-and-wait-evaluation {"model" model-id "dataset" test-id}))
+    (when delete-resources
+          (map safe-delete [training-id test-id model-id]))
```
jaor: indentation

anomaly-benchmarking/make-binary/library.whizzml
```diff
@@ -220,17 +228,19 @@
...
        colored-graph (two-color my-graph my-span-tree)
-       field-id (get (fetch ds) "objective_field")
        color-map (mapify (get colored-graph "nodes") "name" "color")
-       filter (color-filter (get field-id "id") color-map))
-    (create-dataset {"origin_dataset" ds
-                     "new_fields" [{"field" filter
-                                    "name" "Is Anomaly?"}]})))
+       my-filter (color-filter field-id color-map))
+    (when delete-resources
+          (safe-delete my-eval))
```
jaor: indentation

anomaly-benchmarking/make-binary/metadata.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "Make binary",
+  "description": "",
```
jaor: a description would be nice, and perhaps a less generic name

anomaly-benchmarking/make-binary/readme.md
```diff
@@ -0,0 +1,41 @@
+# Making a Dataset Binary
+
+This library takes any dataset and returns it extended by a new field,
+"Is Anomaly?". "Is Anomaly?" takes the values "normal" or "anomalous"
```
jaor: this is just a matter of taste, but i'd name it simply "anomaly?", and, at any rate, define a variable for use by clients of the library so that they don't have to hardcore the string in their code.  also, if the name ends in a question mark, its values should be "true" and "false" or "yes", "no" by usual conventions.  If the values are going to be the adjectives above, i'd use as field name something like "anomaly degree" or somesuch, but "anomaly?"/"true"/"false" is shorter and easier to remember/use.

anomaly-benchmarking/make-binary/readme.md
```diff
@@ -0,0 +1,41 @@
...
+following the procedure laid out in the ODD paper
+http://www.outlier-analytics.org/odd13kdd/papers/emmott,das,dietterich,fern,wong.pdf.
+
+Given a dataset id, the field id of its objective, and the boolean
+`delete-resources`, it:
+
+- Determines whether the dataset is a regression, a classifications
```
jaor: typo: classification

anomaly-benchmarking/metadata.json
```diff
@@ -1,20 +1,9 @@
...
-      }
-  ],
-  "outputs": [
-      {
-          "name": "binary-dataset",
-          "type": "dataset-id",
-          "description": "Extended binary dataset"
-      }
+  "name": "Anomaly benchmarking",
+  "description": "",
```
jaor: missing description

jaor (changes requested):
A few comments in... most of them are just formatting nits.

anomaly-benchmarking/generate-datasets/script.whizzml
akashenfelter: Multiple strings! This will make formatting so much nicer!

anomaly-benchmarking/generate-datasets/script.whizzml
akashenfelter: Sorry about all the indentation errors. It's a combination of me getting used to formatting conventions, and wanting more space as my expressions get longer and longer. I won't let it happen again. At least it's not tabs? :)

anomaly-benchmarking/make-binary/readme.md
akashenfelter: I changed the name to "instance type". The label is so arbitrary (what is "normal" and "anomalous" depends, among other things, on the order of the rows) it feels too much like lying to call it "anomaly?"/"true".

akashenfelter:
Got it! If I didn't comment inline above, I've made the suggested change.  Rebased and ready to go.

anomaly-benchmarking/generate-datasets/script.whizzml
akashenfelter: I've discovered an issue with using "all_fields". When "all_fields" is false, the output dataset has three fields: the old objective field and the two probability fields. When "all_fields" is true, it takes all the fields from the input dataset, including the objective field. So the output dataset ends up with the objective field twice. @jaor, is this expected behavior?

jaor: yes, that's expected. in general, in a batch prediction the input
dataset won't contain the objective,  and if it does, its value doesn't
need to be the same as the predicted one, so having it twice makes sense
(one from the input, which is not used, one the predicted value).

akashenfelter: Ah, thanks. Then I will have to carefully remove the predicted value one.

whizzmler:
> Ah, thanks. Then I will have to carefully remove the predicted value one.

maybe one thing you can do to make that easier is to give it a different
name in the batch prediction request (by default, we use the same name
as the objective)... e.g. something like "predicted species"

akashenfelter:
That's what I came up with! That made it very easy to add to the "all_but" list.

PR #29, opened 2016-11-10
New filter-non-preferred

jaor:
This is a very simple example for a task requested today at our support channel.  Just a ten-liner in our tutorials section.


jaor:
(@akashenfelter may be useful as the base of one of those whizzml 101 capsules)

mmerce:
looks good and works

PR #30, opened 2016-11-10
Remove debug println

charleslparker:
Left this in there after the last PR


jaor:
thanks.  i've cherry-picked this change in my other open PR. closing manually this one.

PR #31, opened 2016-11-16
Improved test and added user-defined dimension categories

akashenfelter:
The new test is broken into two parts, and takes about 25 minutes to run. This meant adding some user-defined inputs. The default is the same as before, but now the categories for point difficulty and semantic variance and the relative frequency levels can all be changed by the user.

### DIFF
--- a/anomaly-benchmarking/generate-datasets/metadata.json
+++ b/anomaly-benchmarking/generate-datasets/metadata.json
@@ -14,25 +14,25 @@
       "name": "difficulty-map",
       "type": "map",
       "default": {"easy": 0.13333333, "medium": 0.33333333, "hard": 0.5, "very hard": 1},
-      "description": "How to categorize a row's point difficulty. This is a map of category levels (e.g. \"medium\") to the maximum allowed value for that level."
+      "description": "Categories for point difficulty"
     },
     {
       "name": "variance-map",
       "type": "map",
       "default":{"high scatter": 0.25, "medium scatter": 0.5, "low scatter": 1, "low clusteredness": 2, "medium clusteredness": 4, "high clusteredness": 1000000},
-      "description": "How to categorize a dataset's semantic variance. This is a map of category levels (e.g. \"low clusteredness\") to the maximum allowed ratio of variance of normal points to variance of anomalous points."
+      "description": "Categories for semantic variance"
     },
     {
       "name": "frequency-list",
       "type": "list",
       "default": [0.001, 0.005, 0.01, 0.05, 0.1],
-      "description": "Which relative frequencies will be generated."
+      "description": "Relative frequency levels"
     },
     {
       "name": "replicates",
       "type": "integer",
       "default": 10,
-      "description": "How many datasets to make at a given point difficulty, variance, and frequency."
+      "description": "How many datasets to make at given categories/levels"
     },
     {
       "name": "delete-resources",
--- a/anomaly-benchmarking/generate-datasets/metadata.json
+++ b/anomaly-benchmarking/generate-datasets/metadata.json
@@ -11,6 +11,30 @@
       "description": "The resource ids of the 'mother' datasets"
     },
     {
+      "name": "diff-map",
+      "type": "map",
+      "default": {"easy": 0.13333333, "medium": 0.33333333, "hard": 0.5, "very hard": 1},
+      "description": "How to categorize a row's point difficulty. This is a map of category levels (e.g. \"medium\") to the maximum allowed value for that level. The default is {\"easy\" 0.13333333 \"medium\" 0.33333333 \"hard\" 0.5 \"very hard\" 1}."
+    },
+    {
+      "name": "var-map",
+      "type": "map",
+      "default":{"high scatter": 0.25, "medium scatter": 0.5, "low scatter": 1, "low clusteredness": 2, "medium clusteredness": 4, "high clusteredness": 1000000},
+      "description": "How to categorize a dataset's semantic variance. This is a map of category levels (e.g. \"low clusteredness\") to the maximum allowed ratio of variance of normal points to variance of anomalous points. The default is {\"high scatter\" 0.25 \"medium scatter\" 0.5 \"low scatter\" 1 \"low clusteredness\" 2 \"medium clusteredness\" 4 \"high clusteredness\" 1000000}."
+    },
+    {
+      "name": "freq-list",
+      "type": "list",
+      "default": [0.001, 0.005, 0.01, 0.05, 0.1],
+      "description": "Which relative frequencies will be generated. The default for this list is [0.001 0.005 0.01 0.05 0.1]."
+    },
+    {
+      "name": "replicates",
+      "type": "integer",
+      "default": 10,
+      "description": "How many datasets to make at a given point difficulty, variance, and frequency. The default is ten."
+    },
+    {
       "name": "delete-resources",
       "type": "boolean",
       "default": true,
--- a/anomaly-benchmarking/generate-datasets/metadata.json
+++ b/anomaly-benchmarking/generate-datasets/metadata.json
@@ -11,28 +11,28 @@
       "description": "The resource ids of the 'mother' datasets"
     },
     {
-      "name": "diff-map",
+      "name": "difficulty-map",
       "type": "map",
       "default": {"easy": 0.13333333, "medium": 0.33333333, "hard": 0.5, "very hard": 1},
-      "description": "How to categorize a row's point difficulty. This is a map of category levels (e.g. \"medium\") to the maximum allowed value for that level. The default is {\"easy\" 0.13333333 \"medium\" 0.33333333 \"hard\" 0.5 \"very hard\" 1}."
+      "description": "How to categorize a row's point difficulty. This is a map of category levels (e.g. \"medium\") to the maximum allowed value for that level."
     },
     {
-      "name": "var-map",
+      "name": "variance-map",
       "type": "map",
       "default":{"high scatter": 0.25, "medium scatter": 0.5, "low scatter": 1, "low clusteredness": 2, "medium clusteredness": 4, "high clusteredness": 1000000},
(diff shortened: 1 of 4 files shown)


anomaly-benchmarking/generate-datasets/metadata.json
```diff
@@ -11,6 +11,30 @@
...
+      "description": "How to categorize a row's point difficulty. This is a map of category levels (e.g. \"medium\") to the maximum allowed value for that level. The default is {\"easy\" 0.13333333 \"medium\" 0.33333333 \"hard\" 0.5 \"very hard\" 1}."
+    },
+    {
+      "name": "var-map",
+      "type": "map",
+      "default":{"high scatter": 0.25, "medium scatter": 0.5, "low scatter": 1, "low clusteredness": 2, "medium clusteredness": 4, "high clusteredness": 1000000},
+      "description": "How to categorize a dataset's semantic variance. This is a map of category levels (e.g. \"low clusteredness\") to the maximum allowed ratio of variance of normal points to variance of anomalous points. The default is {\"high scatter\" 0.25 \"medium scatter\" 0.5 \"low scatter\" 1 \"low clusteredness\" 2 \"medium clusteredness\" 4 \"high clusteredness\" 1000000}."
+    },
+    {
+      "name": "freq-list",
```
jaor: since these are inputs representing the interface of the script with its users, i'd use full names for the parameters, namely variance-map, difficulty-map and frequency-list.

anomaly-benchmarking/generate-datasets/metadata.json
```diff
@@ -11,6 +11,30 @@
...
+      "type": "map",
+      "default": {"easy": 0.13333333, "medium": 0.33333333, "hard": 0.5, "very hard": 1},
+      "description": "How to categorize a row's point difficulty. This is a map of category levels (e.g. \"medium\") to the maximum allowed value for that level. The default is {\"easy\" 0.13333333 \"medium\" 0.33333333 \"hard\" 0.5 \"very hard\" 1}."
+    },
+    {
+      "name": "var-map",
+      "type": "map",
+      "default":{"high scatter": 0.25, "medium scatter": 0.5, "low scatter": 1, "low clusteredness": 2, "medium clusteredness": 4, "high clusteredness": 1000000},
+      "description": "How to categorize a dataset's semantic variance. This is a map of category levels (e.g. \"low clusteredness\") to the maximum allowed ratio of variance of normal points to variance of anomalous points. The default is {\"high scatter\" 0.25 \"medium scatter\" 0.5 \"low scatter\" 1 \"low clusteredness\" 2 \"medium clusteredness\" 4 \"high clusteredness\" 1000000}."
+    },
```
jaor, line 24: in this and the other new parameters, i don't think we need to spell out the default in the description: there's already the "default" field in the metadata for that, which is also shown in the web UI.

anomaly-benchmarking/generate-datasets/script.whizzml
```diff
@@ -1,3 +1,12 @@
+;; Takes a map of bin labels and max edges, and returns a sorted list
+;; of maps
+
+(define (map-to-list my-map)
```
jaor: i wouldn't use such an informative name, my-map, for a function parameter (as an aside, i must confess my- prefixes make me cringe, but it you like them i won't object provided they're kept for naming local variables... that said, it's not something i'll ever sanction in a whizzml style guide :-))  here something like labes-to-edges or simply edges or edges-map would be more adequate.

anomaly-benchmarking/generate-datasets/script.whizzml
```diff
@@ -21,27 +30,37 @@
...
         lr-id (create-and-wait-logisticregression ds)
-        bp-id (wait (create-batchprediction lr-id ds {"probabilities" true 
-                                                      "output_dataset" true 
-                                                      "all_fields" true }))
+        bp-id (wait (create-batchprediction lr-id 
+                                            ds 
+                                            {"probabilities" true 
+                                             "output_dataset" true 
+                                             "all_fields" true 
+                                             "prediction_name" "bp"}))
```
jaor: this will break if an objective field is named "bp", won't it? you could do something like defining `pred-name (str "prediction for " ((fetch ds) ["objective_field" "name"]))` and then use `pred-name` here and below in the `get-id` call.

anomaly-benchmarking/generate-datasets/script.whizzml
```diff
@@ -131,16 +153,16 @@
...
-(define (relative-frequency ds-id normal-rows)
+(define (relative-frequency ds-id normal-rows replicates freq-list)
   (let (ds (fetch ds-id)
         rows (ds "rows")
         ds-name (ds "name")
         my-tags (ds "tags")
         max-freq (/ rows (+ rows normal-rows)))
-    (for (freq [0.001 0.005 0.01 0.05 0.1])
+    (for (freq freq-list)
       (let (k (/ (* freq normal-rows) (- 1 freq)))
```
jaor: can `freq` equal 1?

anomaly-benchmarking/generate-datasets/script.whizzml
```diff
@@ -131,16 +153,16 @@
...
         (when (> max-freq freq)
-          (repeatedly (min 10 (floor (/ rows k))) 
+          (repeatedly (min replicates (floor (/ rows k)))
```
jaor: and can't k equal zero?  (sorry if i already asked back in the day!)

jaor (changes requested):
looks mostly good, just a couple stylistic nits and a potential problem

anomaly-benchmarking/generate-datasets/metadata.json
akashenfelter: I didn't know that! All the "default" description will be removed.

anomaly-benchmarking/generate-datasets/metadata.json
akashenfelter: Good call

anomaly-benchmarking/generate-datasets/script.whizzml
akashenfelter: Naming is one of my greatest weaknesses. I'll change that, and try to avoid my- type names as much as I can.

anomaly-benchmarking/generate-datasets/script.whizzml
akashenfelter: True, will fix.

anomaly-benchmarking/generate-datasets/script.whizzml
akashenfelter: Not sensibly. It is the relative frequency of anomalies, so if it is 1 that means that dataset is all anomalies and no normal points. Should I put in a catch in case someone is trying to use 1 without thinking?

anomaly-benchmarking/generate-datasets/script.whizzml
akashenfelter: `k` is the number of anomalous points, so it would be strange to test an anomaly detector by purposely giving it a dataset without anomalous points. It's not completely ridiculous, however.  I'll put in a catch for the divide by zero error.

akashenfelter: I was hoping to allow the creation of a test dataset without any anomalous points, but it makes both the other dimensions meaningless. (What's the difficulty of a point that doesn't exist? What's the variance of the empty set?) I'll put a minimum value on k (say 10) so that the variance has some good meaning.

anomaly-benchmarking/generate-datasets/metadata.json
osroca: the `default` attribute in an `input` is used to set the value used as default in the execution, it can be a map, but I think that it has to ve a valid JSON object, no?
On the other hand, the "description" of the input is displayed in a tooltip if it is too long and this ones still looks too long even for the tooltip. I would recommend to write such that long definitions in the general description of the script. Charles did something like that for the SMACDown script.

akashenfelter:
So sorry, I will cut the descriptions down so they fit nicely in the
tooltip. I think the "default" is currently a valid JSON object. Is there
some problem I don't see?

osroca:
ah! sorry about my confusion in those default values for the maps, I was thinking that the space in some of the keywords was going to cause an issue in the parser. They are good.
I think that a summary of the great description in the README of the script should be done in the `description` of the `metadata.json`, so when someone  imports the script, that description will be displayed in the interface ...

image

On the other hand, I've imported your script and the input `replicates` uses a wrong type value, it should be `number`.

whizzmler:
yes i guess it won't hurt

whizzmler:
On Thu, Nov 17 2016, Oscar Rovira wrote:

> > @@ -11,6 +11,30 @@
> >        "description": "The resource ids of the 'mother' datasets"
> >      },
> >      {
> > -      "name": "diff-map",
> > -      "type": "map",
> > -      "default": {"easy": 0.13333333, "medium": 0.33333333, "hard": 0.5, "very hard": 1},
> > -      "description": "How to categorize a row's point difficulty. This is a map of category levels (e.g. \"medium\") to the maximum allowed value for that level. The default is {\"easy\" 0.13333333 \"medium\" 0.33333333 \"hard\" 0.5 \"very hard\" 1}."
> > -    },
> > -    {
> > -      "name": "var-map",
> > -      "type": "map",
> > -      "default":{"high scatter": 0.25, "medium scatter": 0.5, "low scatter": 1, "low clusteredness": 2, "medium clusteredness": 4, "high clusteredness": 1000000},
> > -      "description": "How to categorize a dataset's semantic variance. This is a map of category levels (e.g. \"low clusteredness\") to the maximum allowed ratio of variance of normal points to variance of anomalous points. The default is {\"high scatter\" 0.25 \"medium scatter\" 0.5 \"low scatter\" 1 \"low clusteredness\" 2 \"medium clusteredness\" 4 \"high clusteredness\" 1000000}."
> > -    },
> 
> the default attribute in an input is used to set the value used as
> default in the execution, it can be a map, but I think that it has to
> ve a valid JSON object, no?

yes. as far as i can see, the above is valid JSON, isn't it?

akashenfelter:
The description displayed has been expanded, including a link to the readme proper. A `freq` of 1 now has error handling. And `replicates` has an actual JSON type.

PR #35, opened 2016-12-16
New script removing contiguous duplicate keys from an input dataset

jaor:


deduplicate/readme.md
```diff
@@ -0,0 +1,33 @@
...
+once the first one is selected, but the Flatline filter would go on
+for all rows, so there's a pretty good chance that more than one row
+with the same key remains in the filtered dataset.  To get our final
+result, we just keep the first row for each key in the filtered
+dataset, which is easily accomplished via the Flatline filter
+
+    (!= (f "key") (f "key" -1))
+
+So the WhizzML script's job is mostly trivial: generate the Flatline
+expression an request a dataset for each filter in succession and
```
akashenfelter: typo: "... expression, request... succession, and... "

akashenfelter (changes requested):
Just a typo. I did try to figure out how to do this with just one filter, but couldn't make anything nearly as elegant as your two dataset solution.

whizzmler:
typo fixed! yes, i also look for a while for a one filter solution but
didn't come up with anything good enough.  the main problem i found is
not having in flatline any way of storing state as we traverse the
dataset, or to referring to previous values of the new fields.  sooner
or later we are going to have the former (the latter is much harder).

akashenfelter (approved):

akashenfelter:
Merged!

PR #37, opened 2017-01-24
new sliding-window script

cheesinglee:
I wrote this script to help me with a little analysis I was working on, and I figured it would make a good example script. 

This is my first contribution to this repo, so let me know if I've left something out.


akashenfelter:
I'm getting the test error "KO: no dataset found in results", even though whizzml_results.json is where it expects it to be. Something wrong with my setup?

cheesinglee:
What is the output of this command, if you run it from `sliding-window/test`?

`grep -Po 'dataset/[0-9a-f]{24}' cmd/exec/whizzml_results.json|head -1`

akashenfelter:
`usage: grep [-abcDEFGHhIiJLlmnOoqRSsUVvwxZ] [-A num] [-B num] [-C[num]]
	[-e pattern] [-f file] [--binary-files=value] [--color=when]
	[--context[=num]] [--directories=action] [--label] [--line-buffered]
	[--null] [pattern] [file ...]`

cheesinglee:
Looks like the -P flag is not supported in Mac OS's version of grep. Can you see if it works now?

akashenfelter:
Oh, grep.

That works! Merging...

PR #38, opened 2017-02-02
Added same-optype script

akashenfelter:
This script takes a dataset and a source, and updates the source to ensure it has the same optypes as the dataset. This could be useful when frequently running evaluations built on an old dataset with new data that isn't always autodetecting to the correct optypes.

@ashenfad

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -18,7 +18,7 @@
     "model-per-category",
     "model-per-cluster",
     "remove-anomalies",
-    "same-optype",
+    "unify-optype",
     "seeded-best-k",
     "sliding-window",
     "smacdown-branin",
--- a/metadata.json
+++ b/metadata.json
@@ -18,6 +18,7 @@
     "model-per-category",
     "model-per-cluster",
     "remove-anomalies",
+    "same-optype",
     "seeded-best-k",
     "sliding-window",
     "smacdown-branin",
--- a/readme.md
+++ b/readme.md
@@ -44,7 +44,6 @@ By convention, when the artifact is a library, the files are called
   dataset with a hand-picked root node field.
 - `best-k` Scripts and library implementing Pham-Dimov-Nguyen
   algorithm for choosing the best k in k-means clusters.
-- `same-optype` Script matches the field optypes to a given dataset
 - `seeded-best-k` Scripts and library implementing Pham-Dimov-Nguyen
   algorithm for choosing the best k in k-means clusters, with
   user-provided seeds.
@@ -61,6 +60,7 @@ By convention, when the artifact is a library, the files are called
 - `sliding-window` Script that extends a dataset with new fields
   containing row-shifted values from numeric fields. For casting time
   series forecasting as a supervised learning problem.
+- `unify-optype` Script that matches the field optypes to a given dataset
 
 ## How to install
 
--- a/readme.md
+++ b/readme.md
@@ -44,6 +44,7 @@ By convention, when the artifact is a library, the files are called
   dataset with a hand-picked root node field.
 - `best-k` Scripts and library implementing Pham-Dimov-Nguyen
   algorithm for choosing the best k in k-means clusters.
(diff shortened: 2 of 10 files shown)


readme.md
```diff
@@ -44,6 +44,7 @@ By convention, when the artifact is a library, the files are called
   dataset with a hand-picked root node field.
 - `best-k` Scripts and library implementing Pham-Dimov-Nguyen
   algorithm for choosing the best k in k-means clusters.
+- `same-optype` Script matches the field optypes to a given dataset
```
jaor: what about `update-optypes` or `unify-optypes` or even `match-optypes`? all three sound to me a bit better, but then i'm not a native speaker.  also, should that be "Script that matches"?

same-optype/script.whizzml
```diff
@@ -0,0 +1,28 @@
+;; Given a source with incorrect optypes and a dataset with correct
```
jaor: perhaps "with possibly incorrect"? (i presume the script will work even if the types are already fine)

same-optype/script.whizzml
```diff
@@ -0,0 +1,28 @@
...
+
+(define (same-optype source-id dataset-id)
+  (let (ds (fetch dataset-id)
+        ds-ids (keys (ds "fields"))
+        ds-locale (ds "locale")
+        src-fields (resource-fields source-id)
+        src-ids (keys src-fields)
+        ds-fields (map (lambda (x) 
+                         (find-field (ds "fields") 
+                                     (get-in src-fields [x "name"]))) src-ids)
```
jaor: no need for `get-in`, `(src-fields [x "name"])` should do the same.  also, better put `src-ids` in the next line, nicely aligned with `(lambda`.  are we considering the possibility of a field in the source that is not in the dataset? perhaps it was excluded when creating the dataset... that should amount to use only those ids in the dataset's `input_fields`.

jaor (commented):
looks quite good.  a suggestion about naming, and a question about support for excluded fields.

readme.md
akashenfelter: I like `unify-optypes`, I'll make the name changes.

same-optype/script.whizzml
akashenfelter: Yes, I'll make that more clear.

same-optype/script.whizzml
akashenfelter: I haven't considered that... Good point, I'll change it to only update the fields from the dataset's `input_fields`, and also build the final dataset on just those fields.

akashenfelter:
Changes made as discussed.

metadata.json
```diff
@@ -18,6 +18,7 @@
     "model-per-category",
     "model-per-cluster",
     "remove-anomalies",
+    "same-optype",
```
jaor: you forgot to change the name here

whizzmler:
looks good (apart from the wrong name in metadata.json), but the test is
failing for me... this is what i get:

```
[../examples/unify-optype/test allison]$ VERBOSITY=1 ./test.sh 
Testing unify-optype
Removing stale resources (if any)
-------------------------------------------------------
Registering the package script in bigml.io
-------------------------------------------------------
Source source/5894e6e77e0a8d74ba0049f8 created
Dataset dataset/5894e6f07e0a8d74c100b31e created
-------------------------------------------------------
Executing script on dataset
-------------------------------------------------------
Dataset dataset/5894e6f77e0a8d74bb00c1fc created by script
-------------------------------------------------------
Checking results
KO: optype not set properly in results
```

akashenfelter:
Hmmm, can you look at the dataset/5894e6f77e0a8d74bb00c1fc and see what the optype is for field 000000? If it is "text", I need to up my grep game.

whizzmler:
yes, it is text (actually, both fields are text)

akashenfelter:
How odd. Does this version run any better?

whizzmler:
it does, merged!

PR #40, opened 2017-04-21
best-k and low-coverage

akashenfelter:
Low-coverage is a script requested by a customer that removes fields that are beneath some threshold value. 

Extra bonus content: a small variable name change to best-k so bigmler can import it to the gallery, libraries and all.

### DIFF
--- a/low-coverage/readme.md
+++ b/low-coverage/readme.md
@@ -1,0 +1,16 @@
+# Remove Low Coverage Fields
+
+This script removes all the fields that have coverage less than a
+given threshold coverage. For example, if given a dataset of 1000 rows
+and a threshold value of 0.10, all fields with more than 900 missing
+values will be eliminated.
+
+#Inputs
+
+-  The dataset you wish to alter
+
+- A threshold value between 0 and 1
+
+#Output
+
+- The dataset with the low coverage fields removed
--- a/low-coverage/readme.md
+++ b/low-coverage/readme.md
@@ -5,12 +5,14 @@ given threshold coverage. For example, if given a dataset of 1000 rows
 and a threshold value of 0.10, all fields with more than 900 missing
 values will be eliminated.
 
-#Inputs
+# Inputs
 
 -  The dataset you wish to alter
 
-- A threshold value between 0 and 1
+- A threshold value either as an integer, which will be taken as the
+  largest number of missing values permitted, or a percent, which will
+  be taken as the lowest coverage percent permittied.
 
-#Output
+# Output
 
 - The dataset with the low coverage fields removed
--- a/low-coverage/script.whizzml
+++ b/low-coverage/script.whizzml
@@ -9,8 +9,8 @@
         fields-to-keep (filter (lambda (id) 
                           (< (fields [id "summary" "missing_count"]) mc)) 
                         (keys fields)))
-    (if (< threshold 0)
-        (raise "Your threshold is negative."))
+    (when (< threshold 0)
+        (raise "Threshold must be non-negative."))
     (create-dataset {"origin_dataset" dsid
                      "input_fields" fields-to-keep
                      "name" (str name " - only high coverage") })))
--- a/low-coverage/script.whizzml
+++ b/low-coverage/script.whizzml
@@ -1,0 +1,20 @@
+(define (low-coverage dsid threshold)
+  (let (ds (fetch dsid)
+        size (ds "rows")
+        name (ds "name")
+        fields (keys (ds "fields"))
+        missing (map (lambda (x) (ds ["fields" 
+                                          x 
+                                          "summary" 
+                                          "missing_count"])) 
+                         fields)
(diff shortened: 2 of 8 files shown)


low-coverage/script.whizzml
```diff
@@ -0,0 +1,20 @@
...
+        missing (map (lambda (x) (ds ["fields" 
+                                          x 
+                                          "summary" 
+                                          "missing_count"])) 
+                         fields)
+        missing-pairs (iterate (a [] x fields y missing) (append a [x y]))
+
+        low-list (filter (lambda (x) (< (last x) (* (- 1 threshold) size))) 
+                         missing-pairs)
+        fields-to-keep (map (lambda (x) (head x)) low-list))
```
jaor: why 4 iterations? it can be all computed in one go, no?:

```
fields (ds "fields")
mc (* (- 1 threshold) (ds "rows"))
to-keep (filter (lambda (id) (< (fields [id "summary" "missing_count"]) mc))
                (keys fields))
```

low-coverage/readme.md
```diff
@@ -0,0 +1,16 @@
...
+This script removes all the fields that have coverage less than a
+given threshold coverage. For example, if given a dataset of 1000 rows
+and a threshold value of 0.10, all fields with more than 900 missing
+values will be eliminated.
+
+#Inputs
+
+-  The dataset you wish to alter
+
+- A threshold value between 0 and 1
```
jaor: i think perhaps it'd be nice to make this parameter also accept an integer number.  if someone passes a value in [0,1), it's interpreted as a ratio, just as you do know, and if one passes an integer >= 1, it's taken as an absolute number of rows (say, i want to keep those with less than 23 missings)... 0 has the same meaning as a ratio and as a count, and 1 as a ratio is not really useful, is it?

jaor (commented):
A couple of suggestions...

low-coverage/script.whizzml
akashenfelter: Huh, one is much fewer than four. I like this much better.

low-coverage/readme.md
akashenfelter: This is a good idea. I think I have it working (taking the integer to mean the maximum allowed missing and the ratio to mean the minimum percent covered), but I sometimes get turned around so another pair of eyes might be useful.

low-coverage/script.whizzml
```diff
@@ -2,17 +2,13 @@
...
-                                          "summary" 
-                                          "missing_count"])) 
-                         fields)
-        missing-pairs (iterate (a [] x fields y missing) (append a [x y]))
-
-        low-list (filter (lambda (x) (< (last x) (* (- 1 threshold) size))) 
-                         missing-pairs)
-        fields-to-keep (map (lambda (x) (head x)) low-list))
+        fields (ds "fields")
+        mc (if (< threshold 1)
```
jaor: looks good, but i'd add here also a check to ensure threshold is non-negative, `raise`-ing an error (just a string) otherwise.

jaor (changes requested):
another little suggestion...

low-coverage/script.whizzml
akashenfelter: Like this? Or should we forbid negative thresholds altogether?

low-coverage/script.whizzml
```diff
@@ -9,6 +9,8 @@
         fields-to-keep (filter (lambda (id) 
                           (< (fields [id "summary" "missing_count"]) mc)) 
                         (keys fields)))
+    (if (< threshold 0)
+        (raise "Your threshold is negative."))
```
jaor: yes, like this, but it's better to use `when` for one-legged conditonals:

```
(when (< threshold 0)
  (raise "Threshold must be non-negative"))
```

akashenfelter: Change made!

whizzmler:
okay, code looks good, but the test if not working... probably the grep
in there.  i'm getting:

```
$ VERBOSITY=1 ./test.sh 
-------------------------------------------------------
Test for low-coverage script
Testing low-coverage script -------------
KO: low-coverage not set properly in results
```

that's because "outputs" doesn't have to be the first key in the map:

```
$  cat cmd/results/whizzml_results.json 
{"output_resources": [{"last_update": 1493055223633, "variable":
"filtered-dataset", "progress": 0.0, "task": "Queuing job", "code": 1,
"id": "dataset/58fe36f77e0a8d6c310048b1", "state": "queued"}],
"sources": [["script/58fe36d42a834704eb000077", ""]], "results":
["dataset/58fe36f77e0a8d6c310048b1"], "steps": 1139, "result":
"dataset/58fe36f77e0a8d6c310048b1", "outputs": [["filtered-dataset",
"dataset/58fe36f77e0a8d6c310048b1", "dataset"]]}
```

akashenfelter:
Sorry, I really should have caught that on my own.

whizzmler:
it's working now: merged. thanks!

PR #41, opened 2017-04-25
Models-per-category changes

akashenfelter:
The changes suggested by Oscar to improve the script before making it public in the gallery

### DIFF
--- a/model-per-category/create-category-models/script.whizzml
+++ b/model-per-category/create-category-models/script.whizzml
@@ -25,10 +25,10 @@
 ;; Given a parent dataset, a field and one of its categories,
 ;; create a new dataset containing only the instances of the parent
 ;; that have the given category.
-(define (category-dataset origin-id field category name)
+(define (category-dataset origin-id field category ds-name)
   (let (fltr (flatline "(= (f {{field}}) {{category}})"))
     (create-dataset origin-id {"lisp_filter" fltr
-                               "name" (str  name " > " category)})))
+                               "name" (str  ds-name " > " category)})))
 
 ;; Create a dataset and a model for missing values of field
 (define (missings-resources dataset missings field kind params)
--- a/model-per-category/create-category-models/script.whizzml
+++ b/model-per-category/create-category-models/script.whizzml
@@ -27,7 +27,10 @@
 ;; that have the given category.
 (define (category-dataset origin-id field category)
   (let (fltr (flatline "(= (f {{field}}) {{category}})"))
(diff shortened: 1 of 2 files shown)


model-per-category/create-category-models/script.whizzml
```diff
@@ -27,7 +27,10 @@
 ;; that have the given category.
 (define (category-dataset origin-id field category)
   (let (fltr (flatline "(= (f {{field}}) {{category}})"))
-    (create-dataset origin-id {"lisp_filter" fltr})))
+    (create-dataset origin-id {"lisp_filter" fltr
+                               "name" (str ((fetch origin-id) "name")
```
jaor: this function is called for all categories, always with the same dataset, so re-fetching it all the time is  a bit inefficient: we can make this function take a name argument and compute that name only once just before we loop over `category-dataset` below...

jaor (changes requested):
an optimization...

akashenfelter:
Sorry this fell off my radar. Nit picked!

model-per-category/create-category-models/script.whizzml
```diff
@@ -40,7 +41,8 @@
 ;; Final workflow for multiway first split
 (define (multiway-model dataset field objective kind parameters)
   (let ([cs missings id name] (categories dataset field)
-        ds (for (c cs) (category-dataset dataset field c))
+        name ((fetch dataset) "name")
```
jaor: i don't think that's right: your `name` here is shadowing the `name` returned by `(categories ..)`

jaor (changes requested):
i think there's a problem...

akashenfelter:
Better?

whizzmler:
no, i'm afraid not :) you changed the name at the wrong place, it's in
the let before calling category-dataset that it must be changed... we
had:

```clojure
(let ([cs missings id name] (categories dataset field)
      ps ...
  <code that, presumable uses `name` above>))
```

and then you added

```clojure
(let ([cs missings id name] (categories dataset field)
      name ((fetch dataset) "name") ;; <-- this name is shadowing the name in pthe previous line
      ps ...
  <code that, presumable uses `name` above, but now `name` is different>))
```

(the name of the parameter can be `name`, there's no shading there).


-- 
No, no, you're not thinking, you're just being logical.
 -Niels Bohr, physicist (1885-1962)

akashenfelter:
Sorry, I wasn't thinking.

PR #43, opened 2017-05-17
stacked-predictions: better flatline

akashenfelter:
There were some significant problems with my old flatline expression when the defaults weren't used. This should fix that.

### DIFF
--- a/stacked-predictions/script.whizzml
+++ b/stacked-predictions/script.whizzml
@@ -74,4 +74,4 @@
         success? (resource-done? (fetch (wait popular-id))))
     {"models" models "dataset" popular-id "result" success?}))
 
-;;(define popular-dataset ((make-stack training holdout model-list parameters) "dataset"))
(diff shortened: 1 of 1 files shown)


stacked-predictions/script.whizzml
```diff
@@ -66,14 +60,18 @@
...
+                           (keys (resource-fields id)))
         popular-id (create-dataset {"origin_dataset" id 
-                                 "new_fields" [{"field" (popular-filter id obj-id)
+                                 "new_fields" [{"field" (popular-filter id obj-id new-fields)
                                                 "name" "popular"}]})
         success? (resource-done? (fetch (wait popular-id))))
     {"models" models "dataset" popular-id "result" success?}))
 
-(define popular-dataset ((make-stack training holdout model-list parameters) "dataset"))
+;;(define popular-dataset ((make-stack training holdout model-list parameters) "dataset"))
```
jaor: this seems to be commented out by accident?

jaor (changes requested):

PR #44, opened 2017-05-24
use arg not global

seamusabshere:


jaor:
good catch.  merged! thanks a lot!

PR #47, opened 2017-07-06
Stepwise Regression

akashenfelter:
A tweak of the best-first script that implements forward stepwise regression. Also, it now deletes intermediate resources and picks a new test/training split at each step to prevent overfitting.

### DIFF
--- a/stepwise-regression/readme.md
+++ b/stepwise-regression/readme.md
@@ -1,0 +1,21 @@
+# Forward stepwise regression
+
+Script to select the best features for creating a logistic regression,
+using a greedy algorithm:
+
+- Initialize the set `S` of selected features to the empty set
+- Split the dataset into a test and training set.
+- For each feature `f` not in `S`, model and evaluate with feature
+  set `S + f`
+- Greedily select the feature `f'` with the best performance and add
+  it to S
+
+The script takes as inputs the dataset to use and some threshold
+number `d` (zero by default). If the performance is not improved by at
+least `d`, the no new feature is added and the output is returned. For
+example, if an improvement of 1% is desired, the threshold should be
+0.01. The output is a list of the selected features, as field
+identifiers.
+
+To select the best performance, the script uses the metric
+`average_f_measure` in the evaluations it performs.
--- a/stepwise-regression/readme.md
+++ b/stepwise-regression/readme.md
@@ -12,7 +12,7 @@ using a greedy algorithm:
 
 The script takes as inputs the dataset to use and some threshold
 number `d` (zero by default). If the performance is not improved by at
-least `d`, the no new feature is added and the output is returned. For
+least `d`, then no new feature is added and the output is returned. For
 example, if an improvement of 1% is desired, the threshold should be
 0.01. The output is a list of the selected features, as field
 identifiers.
--- a/stepwise-regression/script.whizzml
+++ b/stepwise-regression/script.whizzml
@@ -1,0 +1,97 @@
+;; A simple function to get the max value in a list
+(define (get-max xs) (reduce (lambda (x y) (if (> x y) x y)) (head xs) xs))
+
+;; Get feature names given ids
+(define (feature-names dataset-id ids)
+  (let (fields ((fetch dataset-id) "fields"))
+    (map (lambda (id) (fields [id "name"])) ids)))
+
+;; Create a dataset sample
+(define (sample-dataset ds-id rate oob)
+  (create-and-wait-dataset {"sample_rate" rate
+                            "origin_dataset"  ds-id
+                            "out_of_bag" oob
+                            "seed" "whizzml-example"}))
+
+;; Split a dataset into training and test sets
+(define (split-dataset ds-id rate)
+  (list (sample-dataset ds-id rate false)
+        (sample-dataset ds-id rate true)))
+
+;; Get the default set of input fields for this dataset (all preferred
+;; fields minus the objective field).
+(define (default-inputs dataset-id obj-id)
+  (let (fields ((fetch dataset-id) "fields")
+        fids (keys fields))
+    (filter (lambda (k) (and (fields [k "preferred"] false) (not (= obj-id k))))
+            fids)))
+
+;; Make a list of logistic regressions using the given dataset,
+;; objective field and selected features.  For each lr, add one of the
+;; potential features in potentials to the input feature list.
+
+(define (make-lr dataset-id obj-field selected potentials)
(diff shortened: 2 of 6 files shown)


stepwise-regression/readme.md
```diff
@@ -0,0 +1,21 @@
...
+- Initialize the set `S` of selected features to the empty set
+- Split the dataset into a test and training set.
+- For each feature `f` not in `S`, model and evaluate with feature
+  set `S + f`
+- Greedily select the feature `f'` with the best performance and add
+  it to S
+
+The script takes as inputs the dataset to use and some threshold
+number `d` (zero by default). If the performance is not improved by at
+least `d`, the no new feature is added and the output is returned. For
```
jaor: typo: then no

stepwise-regression/script.whizzml
```diff
@@ -0,0 +1,99 @@
+;; A simple function to get the max value in a list
+(define (get-max xs) (reduce (lambda (x y) (if (> x y) x y)) (head xs) xs))
```
jaor: this is just `(apply max xs)`

stepwise-regression/script.whizzml
```diff
@@ -0,0 +1,99 @@
...
+(define (sample-dataset ds-id rate oob)
+  (create-and-wait-dataset {"sample_rate" rate
+                            "origin_dataset"  ds-id
+                            "out_of_bag" oob
+                            "seed" "whizzml-example"}))
+
+;; Split a dataset into training and test sets
+(define (split-dataset ds-id rate)
+  (list (sample-dataset ds-id rate false)
+        (sample-dataset ds-id rate true)))
```
jaor: instead of `sample-dataset` and `split-dataset` you can just use the standard library function `create-dataset-split`... in your case the call would be: `(create-dataset-split ds-id rate "whizzml-example")`

stepwise-regression/script.whizzml
```diff
@@ -0,0 +1,99 @@
...
+
+;; Get the default set of input fields for this dataset (all preferred
+;; fields minus the objective field).
+(define (default-inputs dataset-id obj-id)
+  (let (fields ((fetch dataset-id) "fields")
+        fids (keys fields))
+    (filter (lambda (k) (and (fields [k "preferred"] false) (not (= obj-id k))))
```
jaor: if a field doesn't have the key "preferred", it's considered preferred, so you want `(fields [k "preferred"] true)`

stepwise-regression/script.whizzml
```diff
@@ -0,0 +1,99 @@
...
+
+;; Make a list of logistic regressions using the given dataset,
+;; objective field and selected features.  For each lr, add one of the
+;; potential features in potentials to the input feature list.
+
+(define (make-lr dataset-id obj-field selected potentials)
+  (let (lr-req {"dataset" dataset-id "objective_field" obj-field}
+        make-req (lambda (fid)
+                   (assoc lr-req "input_fields" (cons fid selected)))
+        all-reqs (map make-req potentials))
```
jaor: just in case you like it better, you could also compute `all-reqs` with a `for`:

```
all-reqs (for (fid potentials)
           {"dataset" dataset-id 
            "objective_field" obj-field 
            "input_fields" (cons fid selected)})
```

(i find it more readable, but feel free to ignore)

stepwise-regression/script.whizzml
```diff
@@ -0,0 +1,99 @@
...
+(define (select-feature test-dataset-id potentials model-ids)
+  (let (eval-req {"dataset" test-dataset-id}
+        make-req (lambda (mid) (assoc eval-req "model" mid))
+        all-reqs (map make-req model-ids)        
+        evs (map fetch (create-and-wait* "evaluation" all-reqs))
+        vs (map (lambda (ev) (ev ["result" "model" "average_f_measure"] 0)) evs)
+        value-map (make-map potentials vs)
+        max-val (get-max vs)
+        choose-best (lambda (id) (and (= max-val (value-map id)) id))
+        _ (delete* (map (lambda (x) (x "resource")) evs)))
```
jaor: you don't need to put this `delete*` in the let varlist with the `_` dummy: just put it in the let body, right before `[(some ...]`

```
(let (eval-req ...
      choose-best ...)
  (delete* (map ...))
  [(some choose-best potentials) max-val])
```

stepwise-regression/script.whizzml
```diff
@@ -0,0 +1,99 @@
...
+(define (select-features dataset-id threshold objective-id)
+  (let (obj-id (get-objective dataset-id objective-id)
+        input-ids (default-inputs dataset-id obj-id))
+    (loop (selected []
+           potentials input-ids
+           f-measure 0)
+      (let (_ (log-info "Splitting dataset...")        
+            splits (split-dataset dataset-id 0.5)
+            train-id (nth splits 0)
+            test-id (nth splits 1)
```
jaor: use destructuring bind:

```
(let ([train-id test-id] (split-dataset ...)
```

stepwise-regression/script.whizzml
```diff
@@ -0,0 +1,99 @@
...
+            _ (log-info "Making logistic regressions...")
+            model-ids (make-lr train-id obj-id selected potentials)
+            _ (log-info "Selecting feature...")
+            next-feat (select-feature test-id potentials model-ids)
+            _ (log-info "Selected feature is " (next-feat 0))
+            diff (- (next-feat 1) (+ f-measure threshold))
+            _ (log-info "Best f-measure is " (next-feat 1))
+            _ (delete* (concat splits model-ids)))
```
jaor: these last two statements (log and delete*) can go in the body of the let, no need for `_`

jaor (changes requested):
a few comments in!

jaor:
ah, @akashenfelter, two other nits: remember to add the new script to the list in the top-level readme (unless you already did and i'm not seeing it :)) and also, i pushed another PR, so i am afraid this one will need a rebase.  thanks!

stepwise-regression/script.whizzml
akashenfelter: That is indeed more readable.

akashenfelter:
Changes made, rebase successful, and added my script to the top-level metadata as well.

jaor:
thanks allison, looks good now.  i had just merged chee sing's PR and that introduced a trivial conflict: i've rebased and fixed it myself, and pushed the result. so, closing this one manually.  just make sure to force-update your master branch to keep it in sync with current examples/master.

PR #48, opened 2017-07-21
Add scripts to increase quadratically qeighted kappa statistic of predictions

hangarr:
These are the scripts to estimate offsets to predictions that will increase the quadratically weighted kappa statistic.  They can be used with any type of predictor.  The optimize-offsets script computes offsets from a training dataset created with the predictor.  The  batch-apply-offsets script applies the offsets to new batch predictions created with the predictor.


jaor:
looks good, merged! thanks a lot!

PR #54, opened 2018-04-11
Bug fix: using the training split in best-first algorithm

jaor:
We were using the full original dataset to train models.  Also using
the standard library to create random splits, instead of our own.

/cc @talvarez

### DIFF
--- a/best-first/script.whizzml
+++ b/best-first/script.whizzml
@@ -13,11 +13,6 @@
                             "out_of_bag" oob
                             "seed" "whizzml-example"}))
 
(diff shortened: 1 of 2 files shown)


best-first/script.whizzml
```diff
@@ -13,11 +13,6 @@
                             "out_of_bag" oob
                             "seed" "whizzml-example"}))
 
-;; Split a dataset into training and test sets
```
mmerce: is sample-dataset used anywhere else or can it be safely removed as well?

PR #55, opened 2018-05-26
find-neighbors: fix: correctly order by distance

jaor:


mmerce:
I've tested and it works, but do we have tests in this repo? How about adding one?

jaor:
@mmerce test added!

jaor:
@mmerce ping

mmerce:
tests pass, sorry about the delay.

PR #56, opened 2018-08-24
Ordinal encoder

kendian:
Request from Francisco. A simple script for ordinal encoding of categorical fields. If classes have a natural order, that can be supplied by the user; otherwise, the assignment of integers to classes is arbitrary.

### DIFF
--- a/ordinal-encoder/metadata.json
+++ b/ordinal-encoder/metadata.json
@@ -1,6 +1,6 @@
 {
     "name": "Ordinal encoder.",
-    "description": "Given a dataset, encodes categorical fields using ordinal ecoding, which uses a single column of integers to represent field classes (levels).",
+    "description": "Given a dataset, encodes categorical fields using ordinal encoding, which uses a single column of integers to represent field classes (levels). It then creates a new dataset, with additional fields containing ordinal encodings of the categorical fields.\n\nIf classes have a known order (such as Like, Somewhat Like, Neutral, Somewhat Dislike, and Dislike), the integer mapping can be supplied; otherwise, integers are assigned to classes arbitrarily.\n\nFor more information, please see the [readme](https://github.com/whizzml/examples/tree/master/ordinal-encoder).",
     "kind": "script",
     "source_code": "script.whizzml",
     "inputs": [
--- a/ordinal-encoder/metadata.json
+++ b/ordinal-encoder/metadata.json
@@ -1,6 +1,6 @@
 {
-    "name": "Encodes categorical features using ordinal ecoding.",
-    "description": "Given a dataset, encodes categorical features using ordinal ecoding. Ordinal encoding uses a single column of integers to represent the feature classes (levels).",
+    "name": "Ordinal encoder.",
+    "description": "Given a dataset, encodes categorical fields using ordinal ecoding, which uses a single column of integers to represent field classes (levels).",
     "kind": "script",
     "source_code": "script.whizzml",
     "inputs": [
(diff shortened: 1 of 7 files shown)


ordinal-encoder/metadata.json
```diff
@@ -0,0 +1,32 @@
...
+            "type": "list",
+            "default": [],
+            "description": "[Optional] A list of field names or identifiers to encode. Non-categorical fields are ignored. If not supplied, or an empty list, all categorical columns are encoded."
+        },
+        {
+            "name": "mappings",
+            "type": "map",
+            "default": {},
+            "description": "[Optional] A map for the encoding of each field. Top-level keys are field names or identifiers. Values are maps of classes to integers. For example, {'col1': {'A': 0, 'B': 1, 'C': 2}}. If no mapping is given for a field, classes are assumed to have no true order and integers are assigned at random. An incomplete mapping for a field causes missing values in the new (ordinal) field."
+        }
```
jaor: classes do have a "natural" order, at least one that we use everywhere else: their order in the summary.categories list of the field descriptor.  we should use that order to assign consecutive integers, if no other is given.

ordinal-encoder/script.whizzml
```diff
@@ -0,0 +1,46 @@
...
+             (make-map classes (range (count classes))))))
+    {"name" (str (field "name") "_ordinal")
+     "field" (get-field-flatline (field "name") mapping)}))
+
+;; Main workflow function
+(define (ordinal-encode dataset-id target-fields mappings)
+  (let (ds (fetch dataset-id)
+        fields (resource-fields ds)
+        fields
+          (reduce (lambda (m k) (assoc-in m [k "id"] k))
```
jaor: silly style nit: `(` should be aligned with `f` if in a new line.  but usually you would do:

```
fields (reduce (lambda (m k) ...)
               fields
               (keys fields))
```

ordinal-encoder/script.whizzml
```diff
@@ -0,0 +1,46 @@
...
+                   (keys fields))
+        fields
+          (if (= 0 (count target-fields))
+              (values fields)
+              (filter (lambda (x) x)
+                      (map (lambda (x) (find-field fields x)) target-fields)))
+        fields (filter categorical-field? fields)
+        new-fields (map (lambda (x) (get-new-field x mappings)) fields)
+        args
+        {"origin_dataset" dataset-id
```
jaor: here alignment is right, but, as above, you'd usually put `{` next to args...

jaor (changes requested):
looks pretty good to me... just added silly indentation nits and a request for not assigning indexes "at random".  oh, and the script needs to be added to the metadata.json at the repo's top level,  as well as the top level readme.md

ordinal-encoder/metadata.json
kendian: yes, "random" is incorrect. it's the order the classes are returned from the field-categories command. is that just hash order? it does not appear to be alphabetic, or the order in which they first appear, etc

osroca (commented):
Please, if we are going to share this in the gallery, we might want to improve the description of this he script in metadata.json with a summary of the description in the readme plus a link to the example readme (description allows markdown)

jaor:
in the summary, it's ordered by *counts*, breaking draws alphabetically.

whizzmler:
(i meant breaking ties, sorry)

ordinal-encoder/metadata.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "Ordinal encoder.",
-    "description": "Given a dataset, encodes categorical fields using ordinal ecoding, which uses a single column of integers to represent field classes (levels).",
+    "description": "Given a dataset, encodes categorical fields using ordinal encoding, which uses a single column of integers to represent field classes (levels). It then creates a new dataset, with additional fields containing ordinal encodings of the categorical fields.\n\nIf classes have a known order (such as Like, Somewhat Like, Neutral, Somewhat Dislike, and Dislike), the integer mapping can be supplied; otherwise, integers are assigned to classes arbitrarily.\n\nFor more information, please see the [readme](https://github.com/whizzml/examples/tree/master/ordinal-encoder).",
```
osroca: Thanks!

jaor (approved):

PR #57, opened 2018-08-28
Additional options

kendian:
Added options to remove original fields or mark them as non-perferred.

### DIFF
--- a/ordinal-encoder/metadata.json
+++ b/ordinal-encoder/metadata.json
@@ -20,6 +20,18 @@
             "type": "map",
             "default": {},
             "description": "[Optional] A map for the encoding of each field. Top-level keys are field names or identifiers. Values are maps of classes to integers. For example, {'col1': {'A': 0, 'B': 1, 'C': 2}}. If no mapping is given for a field, integers are assigned by class count, in descending order (in the case of ties, classes are ordered alphabetically). An incomplete mapping for a field causes missing values in the new (ordinal) field."
+        },
+        {
+            "name": "remove-original?",
+            "type": "boolean",
+            "default": false,
+            "description": "Whether to remove the original, unencoded target fields from the resulting dataset."
+        },
+        {
+            "name": "mark-non-preferred?",
+            "type": "boolean",
+            "default": false,
+            "description": "Whether to mark the original, unencoded target fields as non-preferred in the resulting dataset. Has no effect if remove-original?=true."
         }
     ],
     "outputs": [
--- a/ordinal-encoder/metadata.json
+++ b/ordinal-encoder/metadata.json
@@ -1,5 +1,5 @@
 {
-    "name": "Ordinal encoder.",
+    "name": "Ordinal encoder",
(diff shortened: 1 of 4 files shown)


jaor (changes requested):
the logic looks good, but the test is failing:

```
$ ./test.sh                                                                  cat: cmd/results/output_dataset.csv: No such file or directory
Expected
id,size,code,size_ordinal,code_ordinal
0,XL,C,2,0
1,S,A,1,1
2,M,C,0,0
3,M,B,0,2
4,S,B,1,2
5,L,C,3,0
6,XL,C,2,0
7,M,A,0,1
Actual
```

kendian:
Hmm, the tests all pass for me. I believe that's the last test in the file. In order to test that fields have been properly marked as non-preferred, I resorted to using curl + jq. I suspect you are missing one or both of those. The other tests all rely on bigmler, but I couldn't figure out how to retrieve the metadata for a dataset using bigmler (the fields map). I tried 'bigmler --dataset <ds-id> --no-model, but that doesn't work. Can this be done with bigmler?

jaor:
On Tue, Aug 28 2018, Ken Baldwin wrote:

> Hmm, the tests all pass for me. I believe that's the last test in the file. In order to test that fields have been properly marked as non-preferred, I resorted to using curl + jq. I suspect you are missing one or both
> of those.

  $ which jq
  /usr/bin/jq
   which curl
  /usr/bin/curl
  $

must be some other thing.  the test fails because

    cmd/results/output_dataset.csv

doesn't exist.  that file doesn't look like a metadata file...

whizzmler:
> I tried 'bigmler --dataset --no-model, but that doesn't work. Can this
> be done with bigmler?

if you are creating the dataset using bigmler, the json is stored
locally in the cmd directory, i think.  not sure if it's explained
somewhere in the docs (https://bigmler.readthedocs.io/en/latest/), but
if you look at the output dir it should be easy to find (unless i'm
misremembering).

kendian:
That's unrelated. One of the tests intentionally generates no output (none of the specified fields are categorical). I should be able to silence that by checking if the file exists before cat-ing it. I'm pretty sure the error is from the last test. (You could confirm that by commenting out the last line in test.sh and rerunning.) I'll try again to write that using bigmler. Maybe curl isn't picking up your authentication?

jaor:
ah, to make bigmler store the json, one needs also to pass the flag
`--store`

kendian:
That works! Thanks. I first use bigmler to execute the scripts, which creates the new ds. Adding --store doesn't seem to do anything there. But then I download the dataset using --to-csv to compare the result with expectations. Adding --store to that call does get the json too. Nice

Oddly, the fields are in a different order when returned this way. For example
"100004"
"000002"
"100003"
"000000"
"000001"
Working on how to get jq to sort them by key.

whizzmler:
On Tue, Aug 28 2018, Ken Baldwin wrote:

> That's unrelated. One of the tests intentionally generates no output (none of the specified fields are categorical). I should be able to silence that by checking if the file exists before cat-ing it. I'm pretty sure

no. if i comment the last test, i still get the previous error:

     $ ./test.sh
     cat: cmd/results/output_dataset.csv: No such file or directory
     Expected
     id,size,code,size_ordinal,code_ordinal
     0,XL,C,2,0
     1,S,A,1,1
     2,M,C,0,0
     3,M,B,0,2
     4,S,B,1,2
     5,L,C,3,0
     6,XL,C,2,0
     7,M,A,0,1
     Actual

the scripts exit at the first error.  it never reaches the curl.  you
need to test that without provoking an error.  i don't understand the
logic: output_dataset.csv doesn't exist.  so if what you're checking is
simply that it doesn't exist, you don't need any cat or comparison?

jaor (approved):

PR #58, opened 2018-09-28
adds multi-label prediction scripts and best-first with cross-validation

mmerce:
A new best-first-cv script based on best-first has been added to:
- use cross-validation
- allow other model types to be used
- allow a list of pre-selected fields to be used

the last point allows to iteratively call the script to improve the results. Also, results add the evaluation metric for every new field subset chosen in the process.

The muiti-label scripts need the existing `items-to-features` script and create models for every item and their predictions.

### DIFF
--- a/best-first-cv/script.whizzml
+++ b/best-first-cv/script.whizzml
@@ -63,23 +63,20 @@
                           model-options
                           evaluation-options
                           delete-resources)
-  (log-info "*** " model-options)
   (let (dataset (fetch (k-fold-datasets 0))
         dataset-name (dataset "name" false))
     (check-dataset-objective-id objective-id dataset)
     (let (objective-name (get-objective-name dataset objective-id)
-          _ (log-info "*** pre evaluations")
           evaluations (create-k-evaluations k-fold-datasets
                                             objective-name
                                             dataset-name
                                             model-options
                                             evaluation-options
                                             delete-resources)
-          _ (log-info "*** post evaluations")
           ;; commented out till wintermute's new version
           ;; evaluations-average (create-and-wait-evaluation {"evaluations" evaluations})
           average-phi (get-average-phi evaluations)
-          _ (log-info "*** average phi:" average-phi))
+          _ (log-info "average phi:" average-phi))
       (when delete-resources
         (map safe-delete evaluations))
       average-phi)))
@@ -163,7 +160,6 @@
 ;; Output: (string) Checked objective field ID
 (define (check-dataset-objective-id objective-id dataset)
   (let (fields (dataset "fields" {})
-        _ (log-info "*** objective:" objective-id)
         objective-ids (choosable-objective-ids fields))
     (when (not (member? objective-id objective-ids))
       (raise {"message" (str "Failed to find the objective ID in the dataset"
@@ -251,17 +247,11 @@
 ;; Output: (list) model IDs
 ;;
 (define (create-k-models type multidatasets objective-name model-options)
-  (for (multidataset (reverse multidatasets))
-    (log-info "** type " type)
-    (log-info "** md " multidataset)
-    (log-info "** model-options: " model-options)
-    (log-info "** objective_field: " objective-name)
-    (let (model (create type
-                  (merge {"datasets" multidataset
-                          "objective_field" objective-name}
-                          model-options)))
-      (log-info "** creating model: " model)
-      (wait model))))
+  (wait* (for (multidataset (reverse multidatasets))
+            (create type
+                    (merge {"datasets" multidataset
+                            "objective_field" objective-name}
+                            model-options)))))
 
 
 ;; create-k-evaluations
@@ -298,7 +288,6 @@
         type (if (or (> number-of-models 1)
                      boosting?) "ensemble" "model")
         multidatasets (map last k-fold-pairs)
-        _ (log-info "*** multidatasets" multidatasets)
         models (cond (> number-of-models 1)
                      (create-k-models type
                                       multidatasets
@@ -318,12 +307,10 @@
                                       multidatasets
                                       objective-name
                                       model-options))
-        _ (log-info "*** models" models)
         evaluations (iterate (es []
                               id dataset-ids
                               mid models
                               idx (range 1 (+ 1 (count dataset-ids))))
-                      (log-info "*** evaluating " mid idx)
                       (let (name (str "Evaluation tested with subset "
                                       idx
                                       " of " dataset-name)
@@ -422,7 +409,6 @@
 ;; Output: (boolean) true if successful, false if not
 ;;
 (define (safe-delete id)
-  (log-info "*** deleting id: " id)
   (try (delete id)
        (catch e
          (log-info (str "Error deleting resource " id " ignored"))
--- a/best-first-cv/script.whizzml
+++ b/best-first-cv/script.whizzml
@@ -83,6 +83,12 @@
       average-phi)))
 
 
+(define (safe-create type args)
+  (try
+    (create type args)
+    (catch e
+      (create type args))))
+
 (define (get-average-phi evaluations)
   (/ (iterate (acc 0 ev evaluations)
        (+ acc ((fetch ev) ["result" "model" "average_phi"])))
@@ -193,11 +199,12 @@
 ;;
 (define (create-k-folds dataset-id k-folds)
   (let (k-fold-fn (lambda (x)
-                    (create-dataset {"origin_dataset" dataset-id
-                                     "row_offset" x
-                                     "row_step" k-folds
-                                     "new_fields" [{"name" "k_fold"
-                                                    "field" (str x)}]}))
+                    (safe-create "dataset"
+                                 {"origin_dataset" dataset-id
+                                  "row_offset" x
+                                  "row_step" k-folds
+                                  "new_fields" [{"name" "k_fold"
+                                                 "field" (str x)}]}))
         dataset-ids (map k-fold-fn (range 0 k-folds)))
     (wait* dataset-ids)))
 
@@ -249,10 +256,10 @@
 ;;
 (define (create-k-models type multidatasets objective-name model-options)
   (wait* (for (multidataset multidatasets)
-            (create type
-                    (merge {"datasets" multidataset
-                            "objective_field" objective-name}
-                            model-options)))))
+            (safe-create type
+                         (merge {"datasets" multidataset
+                                 "objective_field" objective-name}
+                                model-options)))))
 
 
 ;; create-k-evaluations
@@ -316,7 +323,10 @@
                                       idx
                                       " of " dataset-name)
                             opts (assoc evaluation-options "name" name))
-                       (append es (create-evaluation id mid opts)))))
+                       (append es (safe-create "evaluation"
+                                               (merge {"dataset" id
+                                                       "model" mid}
+                                                      opts))))))
     (wait* evaluations)
     (when delete-resources
       (map safe-delete models))
--- a/best-first-cv/script.whizzml
+++ b/best-first-cv/script.whizzml
@@ -1,8 +1,7 @@
-(define K-FOLDS 5)
-
 ;; This code will eventually be defined as a library.
 
 (define MODEL_OPTIONS ["balance_objective"
+                       "input_fields"
                        "missing_splits"
                        "pruning"
                        "weight_field"
@@ -21,6 +20,7 @@
                                           "randomize"
                                           "seed"]))
 (define LOGISTIC_OPTIONS ["balance_fields"
+                          "input_fields"
                           "bias"
                           "c"
                           "missing_numerics"
@@ -31,6 +31,7 @@
                           "regularization"
                           "seed"])
 (define EVALUATION_OPTIONS ["sample_rate"
+                            "input_fields"
                             "out_of_bag"
                             "range"
                             "replacement"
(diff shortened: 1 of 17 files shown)


best-first-cv/script.whizzml
```diff
@@ -0,0 +1,561 @@
...
+                                            objective-name
+                                            dataset-name
+                                            model-options
+                                            evaluation-options
+                                            delete-resources)
+          ;; commented out till wintermute's new version
+          ;; evaluations-average (create-and-wait-evaluation {"evaluations" evaluations})
+          average-phi (get-average-phi evaluations)
+          _ (log-info "model options: " model-options)
+          _ (log-info "average phi: " average-phi))
```
jaor: a nit, but these log-info calls can be inside the body of the let, without needing any dummy _ variable assignments

best-first-cv/script.whizzml
```diff
@@ -0,0 +1,561 @@
...
+      (when delete-resources
+        (map safe-delete evaluations))
+      average-phi)))
+
+
+(define (safe-create type args)
+  (try
+    (create type args)
+    (catch e
+      (create type args))))
```
jaor, line 90: i think that most of the time, creating resources fails after a while, so the call to `create` almost never throws an error... it's only later, when `wait` is called (perhaps implicitly) that the exception is raised.  so without explicitly waiting this function is probably not that useful...

best-first-cv/script.whizzml
```diff
@@ -0,0 +1,561 @@
...
+;; Validates that the argument is a resource ID and its type. Raises an error
+;; if otherwise.
+;;
+;; Inputs:
+;;   resource-id: (string) Resource ID
+;;   type: (string) Type of resource
+;;
+;; Output: (string) Checked resource ID
+(define (check-resource-id resource-id type)
+  (when (not (string? resource-id))
```
jaor: there is a whizzml standard procedure `resource-id?` that also checks that the string is well formed, so here you can say: `(when (not (resource-id? resource-id)`

best-first-cv/script.whizzml
```diff
@@ -0,0 +1,561 @@
...
+(define (create-k-folds dataset-id k-folds)
+  (let (k-fold-fn (lambda (x)
+                    (safe-create "dataset"
+                                 {"origin_dataset" dataset-id
+                                  "row_offset" x
+                                  "row_step" k-folds
+                                  "new_fields" [{"name" "k_fold"
+                                                 "field" (str x)}]}))
+        dataset-ids (map k-fold-fn (range 0 k-folds)))
+    (wait* dataset-ids)))
```
jaor, line 209: ah, here we are waiting... what we would need then instead of `safe-create` is something like `safe-create-and-wait` that tries to create the dataset again if the `wait` fails.

best-first-cv/script.whizzml
```diff
@@ -0,0 +1,561 @@
...
+;; Inputs:
+;;   dataset-ids: (list) List of the k-fold dataset IDs
+;;
+;; Output: (list) List of pairs [hold-out dataset, multidataset with the rest]
+;;
+(define (pair-k-folds dataset-ids)
+  (map (lambda(x)
+         [(nth dataset-ids x)
+          (concat (take x dataset-ids)
+          (drop (+ x 1) dataset-ids))])
```
jaor: indentation is off

best-first-cv/script.whizzml
```diff
@@ -0,0 +1,561 @@
...
+       (range 0 (count dataset-ids))))
+
+
+;; select-map-keys
+;;
+;; Filters the keys in a map, keeping only the ones that appear in the list.
```
jaor: `select-keys` is a standard whizzml procedure that does just that

best-first-cv/script.whizzml
```diff
@@ -0,0 +1,561 @@
...
+;;   model-options: (map) Options for the model or ensemble
+;;
+;; Output: (list) model IDs
+;;
+(define (create-k-models type multidatasets objective-name model-options)
+  (wait* (for (multidataset multidatasets)
+            (safe-create type
+                         (merge {"datasets" multidataset
+                                 "objective_field" objective-name}
+                                model-options)))))
```
jaor, line 245: same thing for safe-create/wait... it's the wait that is going to raise an error most of the time, i think...

best-first-cv/script.whizzml
```diff
@@ -0,0 +1,561 @@
...
+;; Inputs:
+;;   id: (resource-id) resource to be deleted
+;;
+;; Output: (boolean) true if successful, false if not
+;;
+(define (safe-delete id)
+  (try (delete id)
+       (catch e
+         (log-info (str "Error deleting resource " id " ignored"))
+         false)))
```
jaor, line 409: silly nit: (log-info ...) already evaluates to false

best-first-cv/script.whizzml
```diff
@@ -0,0 +1,561 @@
...
+                                             potentials)
+              _ (log-info "Next selected feature: " (next-feat-info "feature")))
+          (recur (append selected (next-feat-info "feature"))
+                 (make-complementary potentials [(next-feat-info "feature")])
+                 (cons {"features" (append selected
+                                           (next-feat-info "feature"))
+                              "phi" (next-feat-info "phi")} queue)))))))
+
+;; A simple function to get the max value in a list
+(define (get-max xs) (reduce (lambda (x y) (if (> x y) x y)) (head xs) xs))
```
jaor, line 462: this is just `(apply max xs)`

best-first-cv/script.whizzml
```diff
@@ -0,0 +1,561 @@
...
+
+
+;; Filters the features in one list to be complementary with the ones in
+;; another list
+(define (make-complementary y-list x-list)
+  (iterate (acc y-list x x-list)
+    (filter (lambda (y) (not (= x (y "feature")))) acc)))
```
jaor, line 469: so x-list is  a list of features but y-list is a list of maps that have a "feature" key? if so, maybe we could say it explicitly in the comment... we are doing this, right?:

```
(let (fs (map (lambda (y) (y "feature")) y-list))
  (filter (lambda (x) (not (member? x fs))) x-list))
```

(that's just for my own understanding, feel free to keep your version if you find it clearer)

best-first-cv/script.whizzml
```diff
@@ -0,0 +1,561 @@
...
+(define (feature-names dataset-id ids)
+  (let (fields ((fetch dataset-id) "fields"))
+    (map (lambda (id) (fields [id "name"])) ids)))
+
+;; Get the default set of input fields for this dataset (all preferred
+;; fields minus the objective field).
+(define (default-inputs dataset-id obj-id)
+  (let (fields ((fetch dataset-id) "fields")
+        fids (keys fields))
+    (filter (lambda (k) (and (fields [k "preferred"] false) (not (= obj-id k))))
```
jaor: if a field doesn't have a "preferred" value, the default is `true`, so i think that should be `(fields [k "preferred"] true)` no?

best-first-cv/script.whizzml
```diff
@@ -0,0 +1,561 @@
...
+                   (cons fid selected)))
+    (map make-req potentials)))
+
+;; Given a set of dataset folds, a list of potential features, and a list of
+;; model ids corresponding to those features, select the best
+;; potential feature by performing an evaluation on each model and
+;; returning the feature with the best performance.
+(define (select-feature k-fold-datasets objective-id options candidates potentials)
+  (let (make-attrs (lambda (inp) (merge {"input_fields" inp}
+                                        options))
```
jaor: just double checking: here you're using merge instead of assoc because you want "input_fields" in `options` (if any) to take precedence over `inp`?

best-first-cv/script.whizzml
```diff
@@ -0,0 +1,561 @@
...
+        potentials (map (lambda (x) (x "feature")) potentials)
+        all-reqs (map make-attrs candidates)
+        cvs (map (lambda (e-attrs) (cross-validation k-fold-datasets
+                                                     objective-id
+                                                     e-attrs
+                                                     {}
+                                                     true)) all-reqs)
+        ;;vs (map (lambda (ev) (ev ["result" "model" "average_phi"] 0)) cvs)
+        vs cvs
```
jaor, line 531: is this intentional? vs has a commented out value above...

best-first-cv/script.whizzml
```diff
@@ -0,0 +1,561 @@
...
+;; Retrieves and checks whether the objective field of the given
+;; dataset is categorical, raising an error otherwise.
+(define (get-objective ds-id obj-id)
+  (let (obj-id (if (empty? obj-id)
+                   (dataset-get-objective-id ds-id)
+                   obj-id)
+        otype ((fetch ds-id) ["fields" obj-id "optype"] "missing"))
+    (when (not (= "categorical" otype))
+      (raise (str "The dataset's objective field must be categorical, "
+                  "but is " otype)))
```
jaor, line 549: in the code above, every time we raise, it's a map with "message"/"code", but here it's only a string... if there's no special reason, we can homogenize (one way or the other, no strong opinion: since this is a script and nobody's catching the exception, it can be a string everywhere... hmmm, or are you thinking that in the cases above the code will eventually be in a library? if so, it makes sense to throw there a map with code and here a string, so this would be perfectly fine).

multi-label/ml-batch-prediction/script.whizzml
```diff
@@ -0,0 +1,57 @@
+;; Find field
+(define (find-field-id fields field)
```
jaor: we have a standard procedure, `find-field` that i think already does this

jaor (changes requested):
A few comments below... we'll also need a rebase because i merged another PR.  And, any chances off having a test or two? :)

best-first-cv/script.whizzml
mmerce: I did not add this code, however maybe as there's an implicit wait for the origin resource in case it's not finished, that might be not useless.

best-first-cv/script.whizzml
mmerce: The function does that plus it checks the type of the resource and sends error rmessages if the conditions are not met.

mmerce: Oh, but I see what you mean, the check is improved by using resource-id? ok.

best-first-cv/script.whizzml
mmerce: Ok, but how do we know the arguments that we used to create the dataset? Another reason to store in a variable the body of the original create calls...

best-first-cv/script.whizzml
mmerce: Is it? where? do you want me to add one space in lines 224 and 225?

best-first-cv/script.whizzml
mmerce: yes. Good catch!

best-first-cv/script.whizzml
mmerce: hmm, they should because they are the potential fields to be selected by the algorithm, but I don't think that using assoc would work differently, would it?

best-first-cv/script.whizzml
mmerce: hmm, It's probably the old code that has been commented out, but I'll double check

best-first-cv/script.whizzml
mmerce: yes, The part that had code+message was thought to be a library in the future.

jaor:
On Mon, Oct 01 2018, mmerce wrote:

> I did not add this code, however maybe as there's an implicit wait for
> the origin resource in case it's not finished, that might be not
> useless.

but if the origin is failed, since the origin's creation is not retried,
the second call to create will fail again with the same error.

jaor:
yes, i see that. i'm just saying that instead of using string? you could
use resource-id?, with all other code intact.

jaor:
On Mon, Oct 01 2018, mmerce wrote:

> Ok, but how do we know the arguments that we used to create the
> dataset? Another reason to store in a variable the body of the
> original create calls...

yes, indeed (i thought these datasets were created in this script, so
the arguments couuld be passed around; doing it right requires a
non-trivial rewrite, so probably next version)

jaor:
On Mon, Oct 01 2018, mmerce wrote:

> @mmerce commented on this pull request.
>
> ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
>
> In best-first-cv/script.whizzml:
>
>> +
> +;; pair-k-folds
> +;;
> +;; Builds a list of pairs of hold-out and complementary datasets for all
> +;; the k-fold dataset IDs.
> +;;
> +;; Inputs:
> +;;   dataset-ids: (list) List of the k-fold dataset IDs
> +;;
> +;; Output: (list) List of pairs [hold-out dataset, multidataset with the rest]
> +;;
> +(define (pair-k-folds dataset-ids)
> +  (map (lambda(x)
> +         [(nth dataset-ids x)
> +          (concat (take x dataset-ids)
> +          (drop (+ x 1) dataset-ids))])
>
> Is it? where? do you want me to add one space in lines 224 and 225?

(concat (take x datasset-ids)
        (drop (+ x 1) dataset-ids))

jaor:
i meant the difference between these two:

(assoc {"a" 3} "a" 4) => {"a" 4}
(merge {"a" 4} {"a" 3}) => {"a" 3}

mmerce:
> i meant the difference between these two: (assoc {"a" 3} "a" 4) => {"a" 4} (merge {"a" 4} {"a" 3}) => {"a" 3}

yes, I understood. In general the question would make sense, but in this case, the user tries to find out the input fields, so it's not going to provide them in options, and it she does it will not make any sense, so we can change to assoc just to make sure we overrule any information she provides there.

mmerce:
> On Mon, Oct 01 2018, mmerce wrote: Ok, but how do we know the arguments that we used to create the dataset? Another reason to store in a variable the body of the original create calls...
> yes, indeed (i thought these datasets were created in this script, so the arguments couuld be passed around; doing it right requires a non-trivial rewrite, so probably next version,


And yes, they are created in the script and the arguments could be passed around (which will add extra arguments in some functions I'm afraid only to do so) but in general you cannot count on that. The other solution would be cleaner.

jaor (approved):

PR #60, opened 2018-10-11
Fixes commas in metadata.json files

petersen-poul:


mmerce:
It's correct, thanks!

PR #63, opened 2018-11-06
New Script: Recursive Feature Elimination

unmonoqueteclea:
Recursive Feature Elimination Script along its metadata and tests.

### DIFF
--- a/recursive-feature-elimination/readme.md
+++ b/recursive-feature-elimination/readme.md
@@ -26,8 +26,8 @@ The script takes as inputs:
 
 and yields as output:
 - The dataset with the selected fields
-- Information about the features selected and the evaluation on 
-  each iteration (only if test-dataset id given)
+- Selected fields names and (if test-dataset id given) the evaluation
+  and removed features on each iteration 
 
 One common approach is to execute the script with a very low `n`.
 In this way, observing the evaluation info, we will be able to guess
--- a/recursive-feature-elimination/readme.md
+++ b/recursive-feature-elimination/readme.md
@@ -1,0 +1,39 @@
+# Recursive Feature Elimination
+Script to select the `n` best features for modeling a given dataset, 
+using a recursive algorithm:
+
+
+
+- Initialize the set `S` of selected features to the full initial 
+dataset
+- Initialize `c` as the current number of features in `S`
+- While `c` is greater than `n`:
+  - Create random forest with features of `S`
+  - Obtain importances from random forest
+  - Remove from `S` the least important feature
+  - If test-dataset is defined, evaluate the random forest
+  - Update `c` with current number of features
+
+The script takes as inputs:
+  - the dataset to use
+  - the number of features (that is, dataset fields) to return
+  - the objective field (target)
+  - the test dataset that should be used in the evaluation 
+    If it is not given, the script won't perform evaluations.
+  - the metric that we want to use for the evaluations
+    - Possible metrics: accuracy, average_f_measure, 
+    average_phi, average_precision, and average_recall 
+
+and yields as output:
+- The dataset with the selected fields
+- Information about the features selected and the evaluation on 
+  each iteration (only if test-dataset id given)
+
+One common approach is to execute the script with a very low `n`.
+In this way, observing the evaluation info, we will be able to guess
+the optimal number of features.
+
+**References**:  
+- Guyon, I., Weston, J., Barnhill, S., & Vapnik, V., 
+"Gene selection for cancer classification using support vector machines", 
+Mach. Learn., 46(1-3), 389--422, 2002.
--- a/recursive-feature-elimination/script.whizzml
+++ b/recursive-feature-elimination/script.whizzml
@@ -81,39 +81,6 @@
                                 "model" model}))
 
 
-  ;; get-importances
-  ;;
-  ;; Selects the field importances in an ensemble
-  ;;
-  ;; Inputs:
-  ;;   ensemble: (ensemble) Ensemble information
-  ;;
-  ;; Outputs: (map) Importances by field name
-
-  (define (get-importances ensemble)
-    (let (dist (ensemble "distributions" false)
-          imps (map (lambda (x) (x "importance" false)) dist)
-          m-num (count imps))
-      (iterate (acc {} m-imps imps)
-        (iterate (m-acc acc imp m-imps)
-          (assoc+ m-acc
-                  (imp 0)
-                  (/ (imp 1) m-num))))))
-
-
-  ;; assoc+
-  ;;
-  ;; Adds the value in the key-value pair to the one existing in the accumulator
-  ;;
-  ;; Input:
-  ;;   acc: (map) accumulator map
-  ;;   key: (string) key to accumulate
-  ;;   value: (number) value to be added
-  (define (assoc+ acc key value)
-    (assoc acc key (+ value (acc key 0))))
-
-
-
   ;; recursive-feature-elimination
   ;;
   ;; Applies recursive feature elimination algorithm and returns a map with
@@ -152,14 +119,9 @@
                                 " fields")
                     ;; Create random forest and obtain its importances
                     rf-id (random-forest dataset-id obj-id 0 selected)
-                    imps (get-importances (fetch rf-id))
+                    imps ((fetch rf-id) "importance")
                     ;; Obtaining minimum field importance
-                    min-imp (minimum-importance imps)
-                    removed-field (filter (lambda (n) (= (imps n) min-imp))
-                                          (keys imps))
-                    ;; Obtain first element, just in case there are two ore 
-                    ;; more fields with the minimum importance
-                    removed-field (removed-field 0)
+                    removed-field (min-key (lambda (x) (imps x)) (keys imps))
                     _ (log-info  "Removed-field " 
                                  removed-field ": "
                                  ((feature-names fields [removed-field]) 0))
@@ -173,14 +135,14 @@
                     ev-result (when ev (ev ["result" "model" evaluation-metric]))
                     ;; Iterations evaluation info
                     iterations (when ev {"features" (count selected)
-                              "evaluation"  ev-result})
+					 "evaluation"  ev-result})
                     evaluations (when ev (append evaluations iterations)))
               (recur selected
                       evaluations))))))
 
 ;; response
 ;;
-;; Creates a human-readable response and clean the created resources
+;; Creates a human-readable response and cleans the created resources
 ;;
 ;; Inputs:
 ;;   selected: (list) Selected fields ids
@@ -191,11 +153,11 @@
 ;; Output: (map) Map with selected field names, selected field ids, and 
 ;;         iterations evaluation info (only if test-dataset-id is present)
 (define (response selected selected-names evaluations do-evaluation)
+  ;; Remove created resources
+  (map safe-delete (created-resources))
   (let( evaluations (if  do-evaluation
                          evaluations
                          ["test-dataset-id needed to create evaluations"]))
-    ; Remove all created resources
-    (for (id (created-resources)) (safe-delete id)) 
     {"selected-fields" selected
      "selected-fields-names" selected-names
      "evaluations" evaluations}))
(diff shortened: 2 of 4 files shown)


recursive-feature-elimination/script.whizzml
```diff
@@ -0,0 +1,291 @@
...
+  ;; get-importances
+  ;;
+  ;; Selects the field importances in an ensemble
+  ;;
+  ;; Inputs:
+  ;;   ensemble: (ensemble) Ensemble information
+  ;;
+  ;; Outputs: (map) Importances by field name
+
+  (define (get-importances ensemble)
```
mmerce: This is correct, but since we added an "importances" attribute with a dictionary that contains the importance, the code could be simplified here.

recursive-feature-elimination/script.whizzml
```diff
@@ -0,0 +1,291 @@
...
+                        do-evaluation)
+              (let (_ (log-info "------- New iteration -------")
+                    _ (log-info "Creating random forest with " 
+                                (count selected) 
+                                " fields")
+                    ;; Create random forest and obtain its importances
+                    rf-id (random-forest dataset-id obj-id 0 selected)
+                    imps (get-importances (fetch rf-id))
+                    ;; Obtaining minimum field importance
+                    min-imp (minimum-importance imps)
```
mmerce: I don't see that you use the min-imp for anything other than removing the least important field... couldn't you just look for the last field as ordered by importance and remove that?

whizzml> (define c {"a" 2 "b" 1})
{"a" 2 "b" 1}
whizzml> (min-key (lambda (x) (c x)) (keys c))

recursive-feature-elimination/script.whizzml
```diff
@@ -0,0 +1,291 @@
...
+                    ;; Iterations evaluation info
+                    iterations (when ev {"features" (count selected)
+                              "evaluation"  ev-result})
+                    evaluations (when ev (append evaluations iterations)))
+              (recur selected
+                      evaluations))))))
+
+;; response
+;;
+;; Creates a human-readable response and clean the created resources
```
mmerce: cleans

recursive-feature-elimination/script.whizzml
```diff
@@ -0,0 +1,291 @@
...
+;;   do-evaluation: (boolean) Whether we performed evaluations
+;;
+;; Output: (map) Map with selected field names, selected field ids, and 
+;;         iterations evaluation info (only if test-dataset-id is present)
+(define (response selected selected-names evaluations do-evaluation)
+  (let( evaluations (if  do-evaluation
+                         evaluations
+                         ["test-dataset-id needed to create evaluations"]))
+    ; Remove all created resources
+    (for (id (created-resources)) (safe-delete id))
```
mmerce: just as a suggestion: (map safe-delete (created-resources)) and could be run before the let block

recursive-feature-elimination/script.whizzml
```diff
@@ -0,0 +1,291 @@
...
+;; Outputs: (string) Same objective field id
+;;
+(define (get-objective ds-id obj-id)
+  (let (obj-id (if (empty? obj-id)
+                   (dataset-get-objective-id ds-id)
+                   obj-id)
+        fields (resource-fields ds-id)
+        obj-id ((find-field fields obj-id) "id")
+        otype (fields [obj-id "optype"] "missing"))
+    (when (not (= "categorical" otype))
```
mmerce: is there a reason to limit that to a classification?

mmerce (changes requested):

recursive-feature-elimination/script.whizzml
unmonoqueteclea: Much easier now!

recursive-feature-elimination/script.whizzml
unmonoqueteclea: Didn't know this procedure, very interesting

recursive-feature-elimination/script.whizzml
unmonoqueteclea: Absolutely not, just silly copy paste

unmonoqueteclea:
I have added a new little feature in last commit, please, review it.

recursive-feature-elimination/script.whizzml
```diff
@@ -0,0 +1,248 @@
...
+  ;; deletes resource ignoring errors
+  ;; Inputs:
+  ;;   id: (resource-id) resource to be deleted
+  ;;
+  ;; Output: (boolean) true if successful, false if not
+  ;;
+  (define (safe-delete id)
+    (try (delete id)
+         (catch e
+           (log-info (str "Error deleting resource " id " ignored"))
```
jaor: `log-info` is already a variadic function:

```
(log-info "Error deleting resource " id " ignored")
```

recursive-feature-elimination/script.whizzml
```diff
@@ -0,0 +1,248 @@
...
+           false)))
```
jaor: calls to `log-info` always evaluate to `false`, so this one is not needed

recursive-feature-elimination/script.whizzml
```diff
@@ -0,0 +1,248 @@
...
+                               "objective_field" obj
+                               "randomize" true
+                               "random_candidate_ratio" 0.33
+                               "sample_rate" 0.8
+                               "seed" (str "BigML Boruta" run)
+                               "input_fields" input-fields
+                               "ensemble_sample" {"rate" 1
+                                                  "replacement" true
+                                                  "seed" (str "BigML Boruta" 
+                                                               run) }}))
```
jaor: you're using the same seed as before, which is fine, but then it's much better to define it once in a `let` and then use it in both places, to make the decision obvious.  alternatively, take seed as a parameter rather than n (and then make sure you always send the same seed at the caller location).

recursive-feature-elimination/script.whizzml
```diff
@@ -0,0 +1,248 @@
...
+;;
+;; Inputs:
+;;   model-id: (model-id) Model to evaluate
+;;   test-ds-id: (dataset-id) Test Dataset id
+;; Output:
+;;   evaluation: (evaluation) Result of the model evaluation
+
+  (define (evaluate-model model test-ds-id)
+    (create-and-wait-evaluation {"dataset" test-ds-id
+                                "model" model}))
```
jaor: this function is so trivial that i wouldn't even define it... i'd just call:

```
(wait (create-evaluation test-ds-id model))
```
wherever you're calling `evaluate-model`.  note that it's okay to just use the model and dataset ids directly (and order doesn't matter).

recursive-feature-elimination/script.whizzml
```diff
@@ -0,0 +1,248 @@
...
+              (let (_ (log-info "------- New iteration -------")
+                    _ (log-info "Creating random forest with " 
+                                (count selected) 
+                                " fields")
+                    ;; Create random forest and obtain its importances
+                    rf-id (random-forest dataset-id obj-id 0 selected)
+                    imps ((fetch rf-id) "importance")
+		    ;; Obtain fields with importance 0
+		    ;; (they  doesnt appear in "importance")
+		    no-imp-fields (filter (lambda (n)
```
jaor: unless github is lying to me, indentation is wrong here

recursive-feature-elimination/script.whizzml
```diff
@@ -0,0 +1,248 @@
...
+					    (not (member? n (keys imps))))
+					  selected)
+                    ;; Obtaining minimum field importance
+                    removed-field (min-key (lambda (x) (imps x)) (keys imps))
+		    removed-fields (cons removed-field no-imp-fields)
```
jaor: ditto (i am assuming there are no TABs in the source, right?)

recursive-feature-elimination/script.whizzml
```diff
@@ -0,0 +1,248 @@
...
+                                (count selected) 
+                                " fields after filtering")
+                    ev (when do-evaluation 
+                           (fetch(evaluate-model rf-id test-dataset-id)))
+                    ev-result (when ev (ev ["result" "model" evaluation-metric]))
+                    ;; Iterations evaluation info
+                    iterations (when ev {"features" (count selected)
+					 "evaluation"  ev-result
+					 "last-removed" removed-fields-names})
+                    evaluations (when ev (append evaluations iterations)))
```
jaor: this `let` is quite long, and it's kind of difficult to follow what's going on... any chance you could define some functions to make it shorter? or maybe it's this long because of the many logs: perhaps we don't need that many?

recursive-feature-elimination/script.whizzml
```diff
@@ -0,0 +1,248 @@
...
+;; Returns feature names given ids
+;;
+;; Inputs:
+;;   fieds: (map) Fields map (Output from resource-fields)
+;;   ids: (list) List of field ids
+;;
+;; Output: (list) List with all the field names
+;;
+(define (feature-names fields ids)
+    (map (lambda (id) (fields [id "name"])) ids))
```
jaor: indentation

recursive-feature-elimination/script.whizzml
```diff
@@ -0,0 +1,248 @@
...
+;;
+(define (get-objective ds-id obj-id)
+  (let (obj-id (if (empty? obj-id)
+                   (dataset-get-objective-id ds-id)
+                   obj-id)
+        fields (resource-fields ds-id)
+        obj-id ((find-field fields obj-id) "id")
+        otype (fields [obj-id "optype"] "missing"))
+    (when (= "missing" otype)
+      (raise (str "The dataset's doesn't exist")))
```
jaor: no need for `str`

recursive-feature-elimination/script.whizzml
```diff
@@ -0,0 +1,248 @@
...
+;; Outputs: (string) Same objective field id
+;;
+(define (get-objective ds-id obj-id)
+  (let (obj-id (if (empty? obj-id)
+                   (dataset-get-objective-id ds-id)
+                   obj-id)
+        fields (resource-fields ds-id)
+        obj-id ((find-field fields obj-id) "id")
+        otype (fields [obj-id "optype"] "missing"))
+    (when (= "missing" otype)
```
jaor: no big deal, but it's a bit more idiomatic to use `false` as the not-found value, and then:

```
(when (not otype)
  (raise ...))
```

recursive-feature-elimination/readme.md
```diff
@@ -0,0 +1,39 @@
...
+  - the test dataset that should be used in the evaluation 
+    If it is not given, the script won't perform evaluations.
+  - the metric that we want to use for the evaluations
+    - Possible metrics: accuracy, average_f_measure, 
+    average_phi, average_precision, and average_recall 
+
+and yields as output:
+- The dataset with the selected fields
+- Information about the features selected and the evaluation on 
+  each iteration (only if test-dataset id given)
```
jaor: please document here what that "information" contains.

jaor (changes requested):
looks pretty good, but found some nits, see below.

recursive-feature-elimination/script.whizzml
unmonoqueteclea: In emacs, indentation looked good so, maybe yes, I may have used TAB in some place.
I have run untabify. (And I have written in big letters in my notebook TABS ARE EVIL)

unmonoqueteclea:
Sorry for the strange commit. I tried to remove the blank space that (I don't know why) I had
before each line in previous version.

jaor:
> In emacs, indentation looked good so, maybe yes, I may have used TAB in some place.
> I have run untabify. (And I have written in big letters in my notebook TABS ARE EVIL)

they are!  if you set

```elisp
(setq-default indent-tabs-mode nil)
```

in your config, emacs won't insert any tabs, and you won't need to run
untabify manually (that's already done for you if you have our emacs
config).

jaor (dismissed):

jaor:
@mmerce are you happy with this now? looks good to me

mmerce:
@jaor, looks good to me too. I did not test it, though

mmerce (dismissed):

jaor:
the tests work for me.  now we just need a rebase, @unmonoqueteclea.

jaor (approved):

PR #66, opened 2018-11-14
Default evaluation metrics in case the user doesn't select one

unmonoqueteclea:
Fixes bug when user doesn't select any metric for the evaluation.
Tests are passing.

### DIFF
--- a/recursive-feature-elimination/script.whizzml
+++ b/recursive-feature-elimination/script.whizzml
@@ -135,14 +135,11 @@
 ;;
 ;; Outputs: (string) Chosen evaluation-metric 
 (define (choose-eval-metric evaluation-metric otype)
-  (let (empty-eval (empty? evaluation-metric)
-        classification (= "categorical" otype)
-        eval (if empty-eval
-               (if classification
-                 "average_phi"
-                 "r_squared"
-                 evaluation-metric)))
-    eval))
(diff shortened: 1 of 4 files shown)


unmonoqueteclea:
Sorry, didn't found this bug until today

recursive-feature-elimination/script.whizzml
```diff
@@ -119,8 +121,31 @@
...
+;;
+;; Outputs: (string) Chosen evaluation-metric 
+(define (choose-eval-metric evaluation-metric otype)
+  (let (empty-eval (empty? evaluation-metric)
+        classification (= "categorical" otype)
+        eval (if empty-eval
+               (if classification
+                 "average_phi"
+                 "r_squared"
+                 evaluation-metric)))
```
jaor: i don't think this is doing what you mean, the second `if` has 4 args; usually using `cond` instead of nested `if`s make the code more readable.  also, i wouldn't define separate variables for the conditions, they're too short:

```
eval (cond (empty? evaluation-metric) evaluation-metric
           (= "categorical" optype) "average_phi"
           "r_squared"))
```

although i am not really sure if that's what you where trying to write above (i am guessing not)

jaor (changes requested):

jaor (approved):

PR #67, opened 2018-11-20
New script - name clusters

unmonoqueteclea:
Script to give names to clusters using important field names and their values.


unmonoqueteclea:
Should I create a new cluster instead updating the existing one to achieve immutability or it is fine to update the existing cluster?

name-clusters/script.whizzml
```diff
@@ -0,0 +1,121 @@
...
+;; Returns importances from the model
+;;
+;; Inputs:
+;;   model-id: (map) Corresponding model ID
+;;
+;; Output: (list) String representing the field = value name
+(define (model-importances model-id)
+  (let (model (fetch model-id)
+        importances-ids (model ["model" "importance"]))
+    (map (lambda (n) (head n))
```
jaor: this is just `(map head importances-ids)`

name-clusters/script.whizzml
```diff
@@ -0,0 +1,121 @@
...
+        n (- number-of-terms 1)
+    ;; List of already named clusters to avoid repeating names
+        seen (map (lambda (c) (c "name")) (values clusters)))
+    (loop (cs (tail name-strs))
+     (if (empty? cs)
+        (head name-strs)
+        (let (c (join separator (cons (head name-strs) (take n cs))))
+          (if (member? c seen)
+        (recur (tail cs))
+        c))))))
```
jaor: not sure if it's github, but indentation here seems off.

name-clusters/script.whizzml
```diff
@@ -0,0 +1,121 @@
...
+                      number-of-terms
+                      separator)
+  (let (importances-ids (model-importances model-id)
+        name-strs (map (lambda (x) (str (fields [x "name"] "")
+                                        " = "
+                                        (centroid ["center" x])))
+                       importances-ids)
+        n (- number-of-terms 1)
+    ;; List of already named clusters to avoid repeating names
+        seen (map (lambda (c) (c "name")) (values clusters)))
```
jaor, line 63: you don't update the list of `seen` names in the loop below, as you add new ones: is that right?

name-clusters/script.whizzml
```diff
@@ -0,0 +1,121 @@
...
+;; Output: (list) List of cluster names
+;;
+(define (rename-clusters cl-id number-of-terms separator)
+  (let (cl (fetch cl-id)
+        fields (cl ["clusters" "fields"])
+        models? (cl "model_clusters")
+        centroids (cl ["clusters" "clusters"])
+        centroid-ids (map (lambda (c) (c "id")) centroids)
+        ;; generate models for all centroids
+        model-ids (when models?
```
jaor: instead of logging an error below and fail silently, i'd just raise an exception:

```
_ (or (cl "model_clusters") (raise "Cluster doesn't have any model"))
```

and then you can forget about `models?` and all the logic checking it.

name-clusters/script.whizzml
```diff
@@ -0,0 +1,121 @@
...
+                                                  centroid
+                                                  fields
+                                                  model-id
+                                                  number-of-terms
+                                                  separator)}))))
+    (if models?
+      (prog (try (update cl-id {"clusters" clusters})
+                 (catch e
+                   (log-error "Error updating clusters: " e)
+                   (raise e)))
```
jaor: see above: just raise an error when you discover models? is false.  and the `log-error` in a catch that simply re-raises it is not very useful: the failed execution will contain information about the error already, including a stack trace.

readme.md
```diff
@@ -84,6 +84,10 @@ By convention, when the artifact is a library, the files are called
   uses cross-validation.
 - `multi-label` Classification for datasets with a
    multi-label (items) objective field.
+- `recursive-feature-elimination` Script to select the n best features for 
+   modeling a given dataset, using a recursive algorithm
+- `name-clusters`  Script to give names to clusters using important field
```
jaor, line 89: we also have to add the script to the top-level `metadata.json`... and perhaps here we could list it next to the one that names topic models

jaor (changes requested):

jaor:
(updating the cluster is fine, it's only metadata)

jaor:
i'm thinking that if we want to be totally general, we could take as input a boolean flag (defaulting as false) to tell us whether the cluster should be cloned before changing its names, but only if the addition is more or less trivial (in my experience is not a common use case).

name-clusters/script.whizzml
unmonoqueteclea: I am not sure if I am getting the point, but this loop is inside a function which is executed for each cluster name. On each function execution, seen contains the updated data of previous name clusters.

jaor:
> I am not sure if I am getting the point, but this loop is inside a
> function which is executed for each cluster name. On each function
> execution, seen contains the updated data of previous name clusters.

it was me who wasn't getting the point.  ok!

readme.md
unmonoqueteclea: The one that names topic models is not yet here, am I wrong?

unmonoqueteclea:
> i'm thinking that if we want to be totally general, we could take as input a boolean flag (defaulting as false) to tell us whether the cluster should be cloned before changing its names, but only if the addition is more or less trivial (in my experience is not a common use case).

Ok, I will do it!

jaor (approved):

PR #74, opened 2019-07-17
Automl script

unmonoqueteclea:
First working version of AutoML script. There is a lot more to do in the future but I think this is a good starting point.


automl/feature-selection/script.whizzml
```diff
@@ -52,22 +60,25 @@
   (when (= "dataset" (resource-type train-dataset))
     (create-dataset {"name" (str (resource-name train-dataset) " | filtered")
                      "origin_dataset" train-dataset
-                     "input_fields" selected-fields})))
+                     "input_fields" (field-ids-from-names train-dataset
```
mmerce, line 63: What if the field name is not found in the dataset fields? Isn't that going to cause trouble?

unmonoqueteclea:
Closing this. I am working in a completely new version.

PR #75, opened 2019-07-24
Batch association sets with corrections

mmerce:


batch-association-sets/readme.md
```diff
@@ -0,0 +1,20 @@
...
+the number of rules to be added.
+
+The **inputs** for the script are:
+
+* `dataset-id`: (dataset-id) Dataset ID for the dataset to be extended
+* `association-id`: (association-id) Association ID to be used
+* `dataset-name`: (string) Name for the generated dataset.
+* `rule-ids`: (list) **optional** List of rule IDs to be included in the
+              new dataset
+* `match`: (string) **optional** Condition to be met by the new features
```
unmonoqueteclea, line 16: And what happens if the user doesn't specify a _match_ string? maybe, we should say something about this

batch-association-sets/readme.md
```diff
@@ -0,0 +1,20 @@
...
+user-provided dataset. Each rule will be used to create a new feature that
+contains whether the row information fits into the rule. The user can select
+the number of rules to be added.
+
+The **inputs** for the script are:
+
+* `dataset-id`: (dataset-id) Dataset ID for the dataset to be extended
+* `association-id`: (association-id) Association ID to be used
+* `dataset-name`: (string) Name for the generated dataset.
+* `rule-ids`: (list) **optional** List of rule IDs to be included in the
```
unmonoqueteclea, line 14: if you input an empty list, all the rules will be used, no?

correlations-matrix/metadata.json
```diff
@@ -0,0 +1,38 @@
+{
+  "name": "Correlations matrix",
+  "description": "Creates a matrix of correlations between the numeric and categorical fields in a dataset. The index used by default to measure correlation is **eta squared** (one-way ANOVA) and ",
```
unmonoqueteclea: something missing after the final **and "**?

correlations-matrix/readme.md
```diff
@@ -0,0 +1,18 @@
...
+that contains a matrix of correlation values for the fields in a dataset.
+The index used for the
+correlation analysis is the `eta squared` (One-Way ANOVA). Given that this
+correlation can be computed for numeric and categorical fields, only
+the fields that fall into these types are used.
+
+The **inputs** for the script are:
+
+* `dataset-id`: (dataset-id) Dataset ID for the original data
+* `fields`: (list) List of fields to be analyzed (default: [])
```
unmonoqueteclea: All fields will be analyzed by default...

correlations-matrix/readme.md
```diff
@@ -0,0 +1,18 @@
...
+* `options`:  (map) Arguments for the correlation creation (default: {})
+* `correlation-index`: (string) Index to be used (default: "eta_square")
```
unmonoqueteclea, line 17: which are other possible indexes that can be used?

correlations-matrix/script.whizzml
```diff
@@ -0,0 +1,64 @@
...
+;; Inputs:
+;;   dataset-id: (string) Dataset ID
+;;   s-fields: (list) List of fields to be analyzed
+;;   options: (map) Arguments for the correlations creation
+;;   c-index: (string) Index to be used to analyze correlation
+
+(define (correlations-cross-tabs dataset-id s-fields options c-index)
+  (let (ds (fetch dataset-id)
+        fields (resource-fields ds)
+        find-field-l (lambda (f) (find-field fields f))
```
unmonoqueteclea, line 13: In these cases, I miss **partial functions** in whizzml...
Maybe, it could be a good new primitive, @jaor

supervised-model-quality/script.whizzml
```diff
@@ -0,0 +1,51 @@
+(define SUPERVISED-MODELS ["model", "ensemble", "logistic-regression",
+                           "linear-regression", "deepnet"])
```
unmonoqueteclea, line 2: and fusions?

supervised-model-quality/readme.md
```diff
@@ -0,0 +1,18 @@
+# Supervised model quality
+
+In this package you'll find a script that generates an evaluation from
+an existing model by retrieving the dataset used for training and doing
+an 80%-20% split of the data. The output of the script is a map with the
```
unmonoqueteclea, line 5: If I understood well, we are retraining the model with a subset of the original data and testing it with the other subset, aren't we?  Maybe the user can think that we are evaluating the same model with a subset of the training data...

Anyways, I am testing it and I am having some problems:


ERROR: Problem while executing script: Function expects 7 arguments, 5 given

PROBLEM FOUND:
	with push-procedure at line 29 column 24

STACK TRACE:
	at line 29, column 24
	at line 29, column 25
	at line 51, column 21

unmonoqueteclea (changes requested):
Mainly nits.
But I am not able to execute supervised-model-quality script

whizzmler:
On Thu, Jul 25 2019, unmonoqueteclea wrote:

> +(define (correlations-cross-tabs dataset-id s-fields options c-index)
> +  (let (ds (fetch dataset-id)
> +        fields (resource-fields ds)
> +        find-field-l (lambda (f) (find-field fields f))
>
> In these cases, I miss partial functions in whizzml...
> Maybe, it could be a good new primitive, @jaor

yes. or automatic currying, like haskell, so that calling a function
with less arguments than it takes returns the partial one, so simply
`(find-fields fields)` would be it. with maps as functions we kind of
have it already: `({"a" {"b" 3}} "a")` can be used as a function.

correlations-matrix/readme.md
```diff
@@ -0,0 +1,18 @@
+# Correlation matrix CSV
```
unmonoqueteclea, line 1: Strictly, we are not generating a CSV anywhere.
In fact, how could a user automatically create a csv from correlations-tab output?

unmonoqueteclea (changes requested):

supervised-model-quality/readme.md
mmerce: This problem should be solved now. The function I used to retrieve the model configuration had changed and new arguments were needed. Thanks for the heads up.

supervised-model-quality/script.whizzml
mmerce: Fusions were deliverately left out because, as they mix different models, there's no guarantee that they all have a unique dataset to start from. Thus, there's not a clear way of doing a train/test split for all of them in general.

correlations-matrix/readme.md
mmerce: I asked the backend team about that, and the only correlation index that they suggested was eta_square. However, they mentioned that other indexes could be easily added to the correlation resource that could be also used in this kind of matrices. That's why I offered that index as a parameter, just in case we ended up offering other indexes.

batch-association-sets/readme.md
mmerce: done

batch-association-sets/readme.md
mmerce: yes

correlations-matrix/metadata.json
mmerce: no, I remove the text

correlations-matrix/readme.md
mmerce: yes

correlations-matrix/script.whizzml
mmerce: Yes, I see this pattern so many times...

correlations-matrix/readme.md
mmerce: Yes, that's a matrix currently. I think it might be more useful if we finally output a CSV, so I'll try to change that.

mmerce: Now that I changed it, I remember why I used the matrix finally. When we create a text as output, I don't see how to download the entire contents of it. I only see it truncated. @osroca ?

unmonoqueteclea (dismissed):
The supervised-model-quality script now works fine with everything I tested but deepnets.

I tested it with a deepnet and it gives me this error.

**ERROR: Wrong type of some input values: Errors: [["res-id" "supervised-model-id" "deepnet/5d431203c984172452020250"]]**


It doesn't seem to be your fault. Last week I added to the whizzml input checker of wintermute
"deepnet" and "linearregression" as valid values for supervised-model-id.

But it is not deployed yet, I think everything will work fine when it is deployed

mmerce:
yes, I've tried to do a simple script that receives a supervised-model-id and when I use a deepnet ID that error appears

osroca:
@mmerce @unmonoqueteclea are you trying in production? it has been a while since the last deploy of a new wintermute there and if Pablo say that he added deepnets and linearregressions to the 'subpervised-model-id' "meta-resource" last week, maybe it is not available yet in production ...
The newest wintermute should be in ozone if you want to give it a try

unmonoqueteclea (approved):
Everything is ok now, shall I merge?

mmerce:
yes, please

PR #78, opened 2019-08-07
Bulk move of previously selected resources to a Project

mmerce:


bulk-move/metadata.json
```diff
@@ -0,0 +1,50 @@
+{
+  "name": "Bulk Move of Resources",
+  "description": "Moving the selected resources to a destination Project. The first parameter is the destination project, and the rest of parameters are used to filter the resources to be lected. All provided conditions are applied to the filter. ",
+  "kind": "script",
```
jaor, line 4: typo: lected. (and perhaps Project -> project)

bulk-move/metadata.json
```diff
@@ -0,0 +1,50 @@
...
+  "source_code": "script.whizzml",
+  "inputs": [
+    {
+          "name": "destination-id",
+          "type": "string",
+          "default": "",
+          "description": "Project ID used as destination to move the resources. If empty, a new project named \"Unknown\" will be created and the selected resources will be moved there."
+     },
```
jaor, line 12: i would instead set `default` to "Unknown" (as a plus, no explanation will be needed in the docstring then)

bulk-move/metadata.json
```diff
@@ -0,0 +1,50 @@
...
+    {
+          "name": "project-id",
```
jaor: perhaps "source-id" to mirror "destination-id"? (or source-project/destination-project)

bulk-move/script.whizzml
```diff
@@ -0,0 +1,114 @@
+;; Script to move a bunch of resources selected by user-provided filters
+;; to a destination Project
+
+;; Auxiliary functions to select resources based on usual filters
+;; -- to become a library
+
+(define RES_LISTERS {"fusion" list-fusions
```
jaor: i forgot: isn't the convention "RES-LISTERS"?

bulk-move/script.whizzml
```diff
@@ -0,0 +1,114 @@
...
+                     "anomalyscore" list-anomalyscores
+                     "associationset" list-associationsets
+                     "projection" list-projections
+                     "batchprediction" list-batchpredictions
+                     "batchcentroid" list-batchcentroids
+                     "batchanomalyscore" list-batchanomalyscores
+                     "batchtopicdistribution" list-batchtopicdistributions
+                     "batchprojections" list-batchprojections
+                     "script" list-scripts
+                     "library" list-libraries})
```
jaor: are you sure we need this? there's the standard procedure `resources`, and, for instance `(resources "anomaly" x)` is equivalent to `(list-anomalies x)`...

bulk-move/script.whizzml
```diff
@@ -0,0 +1,114 @@
...
+      (= "" project-id)))
+
+(define (get-list-fns resource-types remove-project)
+  (if (empty? resource-types)
+      (let (listers RES_LISTERS
+            listers (if remove-project
+                        (dissoc listers "project")
+                        listers))
+        (values listers))
+      (filter (lambda (x) (procedure? x))
```
jaor: `(filter procedure? (for ..))`

bulk-move/metadata.json
jaor: oh, and in our next release (coming soon) that can be type "project-id" with a default that is not a project-id, so i'ld go for it right away.

bulk-move/metadata.json
```diff
@@ -0,0 +1,50 @@
...
+          "type": "string",
```
jaor, line 15: and, as before, we can use here "project-id"

bulk-move/script.whizzml
```diff
@@ -0,0 +1,114 @@
...
+
+(define (bulk-move destination-id
+                   project-id
+                   tags
+                   filter-expr
+                   resource-types
+                   dry-run)
+  (try (when (not (check-project-id project-id))
+         (raise "Please, check the project ID"))
+    (catch e (raise "Please, check the project ID")))
```
jaor: `check-project-id` never raises an error, so this seems equivalent to the sorter:

```whizzml
(or (check-project-id project-id)
    (raise "Please, check the project ID"))
```

bulk-move/script.whizzml
```diff
@@ -0,0 +1,114 @@
...
+  (try (when (not (check-project-id destination-id))
+         (raise "Please, check the destination project ID"))
+    (catch e (raise "Please, check the destination project ID")))
```
jaor: same here, no need for try/catch

bulk-move/script.whizzml
```diff
@@ -0,0 +1,114 @@
...
+                             resource-types)
+        counter (iterate (counter {} id ids)
+                  (assoc counter (resource-type id)
+                                 (+ (counter (resource-type id) 0) 1))))
+    (log-info "Selected Resources for: ")
+    (log-info "  " expr)
+    (for (type (keys counter))
+      (log-info "  " type ": " (counter type)))
+    (if dry-run
+        (log-info ids)
```
jaor, line 82: not sure, but maybe we could log also for non-dry runs, no?

jaor (changes requested):

bulk-move/metadata.json
mmerce: done

bulk-move/metadata.json
mmerce: I wanted it to be usable just in case we share with the customer. He seems to prefer using the dashboard and that might take a while.

bulk-move/metadata.json
mmerce: source-id could mislead, right? better source-project and destination-project

bulk-move/script.whizzml
mmerce: I was doubting too, and I guess we've used both but we can decide that is - also for constants.

bulk-move/script.whizzml
mmerce: Oh, I did not know! Sure it's better to use (resources ...)

bulk-move/script.whizzml
mmerce: right

bulk-move/script.whizzml
mmerce: Oh, I thought it did when the project was not an ID.

bulk-move/script.whizzml
mmerce: I do log the summary part.

mmerce:
I think I added most suggestions. Please, let me know if anything is missing.

jaor (approved):

PR #82, opened 2019-08-09
New configuration parameters

unmonoqueteclea:
This PR adds a new configuration input parameter with some parameters that
let the user exclude some fields or models or filter the features that are added 
to the original dataset from the unsupervised models.

In the near future, some new configuration parameters may be added for other
unsupervised models.

### DIFF
--- a/automl/automl-library/library.whizzml
+++ b/automl/automl-library/library.whizzml
@@ -97,6 +97,15 @@
                 " the training execution) or provide a training dataset to start"
                 " training again with the new excluded_fields"))))
 
+;; Checks if excluded-models are valid unsupervised models
+(define (check-excluded-models excluded-models)
+  (let (all-models #["cluster" "anomaly" "association" "pca" "topicmodel"]
+        bads (difference (set* excluded-models) all-models))
+    (when (not (empty? bads))
+      (log-warn "WARNING: The following unsupervised models from "
+                "excluded-models are not valid: "
+                (str (list* bads))))))
+
 ;; Checks if train validation and test datasets contain the same fields
 (define (check-datasets-fields train valid test)
   (let (fields-set (lambda (ds)
@@ -121,6 +130,12 @@
 ;;;;;;;;;;;;; Creation of unsupervised models ;;;;;;;;;;;;;;;;;
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
+;; Whether the model "type" should be created based on the
+;; list of excluded models
+(define (create-unsupervised? type excluded-models)
+  (when (string? type)
+    (not (member? type (or excluded-models [])))))
+
 (define (_set-unsupervised-params dataset params exclude)
   (let (name (str (resource-name dataset) " | unsupervised-gen")
         params (assoc (or params {}) "name" name)
@@ -131,10 +146,11 @@
       params)))
 
 ;; Creates an unsupervised model (specified by type) from a given
-;; dataset and some params. exclude is a flag to exclude
-;; objective id for the model creation
-(define (create-unsupervised type dataset params exclude)
-  (let (final-params (_set-unsupervised-params dataset params exclude))
+;; dataset and some params. Objective field is excluded from all
+;; models but associations
+(define (create-unsupervised type dataset params)
+  (let (exclude (not (= type "association"))
+        final-params (_set-unsupervised-params dataset params exclude))
     (try
      (log-info " - Creating " type)
      (create type dataset final-params)
--- a/automl/automl-library/library.whizzml
+++ b/automl/automl-library/library.whizzml
@@ -83,25 +83,23 @@
     (raise (str "If you don't define a correct train-dataset,"
                 "you must define a correct previous execution"))))
 
-;; Checks if the values of the configuration inputs from
-;; the given automl-execution are different from the ones
-;; given by the user in the current execution
-;; If so, values from automl-execution will be used
-(define (check-config-input new-input prev-input name)
-  (when (not (= new-input prev-input))
-    (log-warn (str name " input is overwritten by the one from the "
-                   "previous execution with value "
-                   prev-input))))
-
+;; Checks that all the needed params are in the configuration
+(define (check-configuration-params config)
+  (let (correct #["excluded-fields" "excluded-models"
+                  "pca-variance-threshold" "leverage-threshold"]
+        dif (difference correct (set* (keys (or config {})))))
+    (when (not (empty? dif))
+      (raise (str "Can't find the following parameters in the "
+                  "configuration-params: " dif)))))
 
 ;; Checks if excluded-models are valid unsupervised models
 (define (check-excluded-models excluded-models)
   (let (all-models #["cluster" "anomaly" "association" "pca" "topicmodel"]
         bads (difference (set* excluded-models) all-models))
     (when (not (empty? bads))
-      (log-warn "WARNING: The following unsupervised models from "
-                "excluded-models are not valid: "
-                (str (list* bads))))))
+      (log-warn (str "WARNING: The following unsupervised models from "
+                     "excluded-models are not valid: "
+                     (list* bads))))))
 
 ;; Checks if train validation and test datasets contain the same fields
 (define (check-datasets-fields train valid test)
@@ -122,7 +120,6 @@
                 " and not in your validation or test datasets (besides the"
                 " objective field)"))))
 
-
 (define (_check-number value min-value max-value name)
   (when (or (not (number? value))
             (not (<= min-value value max-value)))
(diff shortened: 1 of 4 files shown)


automl/automl-library/library.whizzml
```diff
@@ -83,19 +83,23 @@
...
-    (raise (str "Your excluded_fields are different to the ones you provided"
-                " in the given previous execution of automl."
-                " Please, provide the same excluded_fields (or leave blank the"
-                " excluded_fields input to use the same excluded_fields as in"
-                " the training execution) or provide a training dataset to start"
-                " training again with the new excluded_fields"))))
+;; Checks that all the needed params are in the configuration
+(define (check-configuration-params config)
+  (let (correct #["excluded-fields" "excluded-models"
+                  "pca-variance-threshold" "leverage-threshold"]
```
mmerce: leverage-threshold applies to lift too? maybe association-threshold or rule-strenght-threshold?...

automl/automl-library/library.whizzml
```diff
@@ -114,13 +118,32 @@
                  (> (count (difference train-fields test-fields)) 1)))
       (log-warn "WARNING: There are some fields that are in your train dataset"
                 " and not in your validation or test datasets (besides the"
-                " objective field"))))
+                " objective field)"))))
+
+(define (_check-number value min-value max-value name)
```
mmerce: maybe _check-in-range

automl/automl-library/library.whizzml
```diff
@@ -242,29 +266,33 @@
           (lambda (i) (_flatline-item (items i) (fields ((items i) "field_id"))))
           lhs (map get-item-fn (rule "lhs"))
           rhs (map get-item-fn (rule "rhs"))
+          metric (rule "leverage" 1)
```
mmerce: I thought you also used rules based on lift, but maybe you don't

automl/automl-library/library.whizzml
```diff
@@ -242,29 +266,33 @@
...
+              (str "(and " lhs " " rhs ")"))))))
+
+(define (_batch-association-sets dataset-id association-id match threshold)
   (let (association (fetch association-id)
         associations (association "associations" {})
         rules (associations "rules" [])
-        rule-ids (if (empty? (head rule-ids []))
-                   (map (lambda (r) (r "id")) rules)
-                   (head rule-ids))
+        rule-ids (map (lambda (r) (r "id")) rules)
```
mmerce: If you don't filter by the list of rule-ids provided as parameter in the function, then these two following sentences are useless, right?

automl/automl-script/readme.md
```diff
@@ -42,9 +42,29 @@ The **inputs** for the script are:
   `automl-execution` should be provided for the script to work. When
   both are provided, the `automl-execution` argument is discarded and
   new resources are generated from the training data.
-* `excluded-fields`: (list) List of fields that will be excluded from
-  any dataset before any other process starts. e.g. ["bmi",
-  "age"]. Empty by default
+* `configuration-params`: (map) Execution configuration
+  parameters. They will be overwritten if automl-execution is
+  given. Overwritten if automl-execution is given. The following
```
mmerce: Duplicated "overwritten if automl-execution is given".

automl/automl-script/readme.md
```diff
@@ -42,9 +42,29 @@ The **inputs** for the script are:
...
+  * `pca-variance-threshold`: (number) The PCA projection uses the
+  minimum number of components such that the cumulative explained
+  variance is greater than the given threshold. Values from 0 to 1.
+  Default value is 1 (all the components will be used)
+  * `leverage-threshold`: (number) Associations with an absolute
+  leverage lower than the threshold will be ignored.  Default value is
+  0 (all the association will be added to the extended dataset)
+
+
+**WARNING** Remember that, to avoid confusion, `configuration-params`
```
mmerce, line 66: I understand that excluded-fields will  not be used when the automl-execution is given, but the rest of parameters can be used, right?

automl/automl-script/script.whizzml
```diff
@@ -1,61 +1,114 @@
...
+                                             model-list
+                                             name
+                                             model-type
+                                             batch-type
+                                             params))
+        cluster-fields (feat-gen "cluster" "batchcentroid" {})
+        anomaly-fields (feat-gen "anomaly" "batchanomalyscore" {})
+        topic-fields (feat-gen "topicmodel" "batchtopicdistribution" {})
+        pca-params {"variance_threshold" pca-threshold}
+        pca-fields (feat-gen "pca""batchprojection" pca-params)
```
mmerce: space missings

mmerce (changes requested):

mmerce:
mostly style nits. Still untested.

automl/automl-library/library.whizzml
unmonoqueteclea: Leverage and lift take different range of values. Currently I am only filtering by leverage because is the only parameter that has defined values range. Lift can't take all real values. I still create two associations ordering by lift and leverage but I filter both list of rules by leverage. It's not the perfect solution but I can't imagine a better one.

unmonoqueteclea: Anyways, I could change the name as you are suggesting, providing in that way more abstraction.

automl/automl-library/library.whizzml
unmonoqueteclea: Done!

automl/automl-library/library.whizzml
unmonoqueteclea: I have the problem I mentioned previously with lift. It doesn't take a fixed range of values

automl/automl-library/library.whizzml
unmonoqueteclea: Yes, I didn't noticed!

automl/automl-script/readme.md
unmonoqueteclea: Removed!

automl/automl-script/readme.md
unmonoqueteclea: Not in my implementation. **All** the configuration params are inherited from the execution, if it is given. I  have thought about this and I took this decision because I think that it would be difficult to the user to distinguish between parameters that are inherited and parameter that are not. If we take the full automl as a new kind of model (as it could be a deepnet or a single optiml), these are parameters for **training** the model. If we pass a previous automl-execution is because we don't want to train, we only want to evaluate or predict, so we shouldn't change the configuration of the model without training it again. In fact, in this case, even the OptiML is inherated from the given automl-execution. If we change the configuration, e.g. of the execluded-models, and don't train a new OptiML we will have a lot of missing inputs in our prediction

unmonoqueteclea: I would like to know your opinion about this

automl/automl-script/script.whizzml
unmonoqueteclea: Sorry, where?

automl/automl-script/readme.md
mmerce: I was thinking that if I have an existing automl-execution I might want to change the pca-variance-threshold or the leverage-threhsold and run another one. These are like "batch-prediction" config parameters, and can be reset using the same extended dataset and models.

mmerce: After our discussion, I see clearly that you are only using the execution as pointer to the final fusion model to create batch predictions based on the models as they have been built. In that sense, it's totally correct to assume that these parameters will not be used when the execution is provided.

automl/automl-script/script.whizzml
mmerce: between "pca" and "batchprojection"

automl/automl-library/library.whizzml
```diff
@@ -120,17 +120,17 @@
...
+                " must be a number "
                 " in the interval [" min-value  ","  max-value  "]"))))
 
 ;; Checks params values
-(define (check-params pca-variance-threshold leverage-threshold)
-  (_check-number pca-variance-threshold 0 1 "pca-variance-threshold")
-  (_check-number leverage-threshold 0 1 "leverage-threshold"))
+(define (check-params pca-variance-threshold max-rules)
+  (_check-in-range pca-variance-threshold 0 1 "pca-variance-threshold")
+  (_check-in-range max-rules 0 1000 "max-rules"))
```
mmerce: maybe getting the integer part of max-rules, just in case....

automl/automl-script/readme.md
```diff
@@ -57,10 +56,12 @@ The **inputs** for the script are:
   minimum number of components such that the cumulative explained
   variance is greater than the given threshold. Values from 0 to 1.
   Default value is 1 (all the components will be used)
-  * `leverage-threshold`: (number) Associations with an absolute
-  leverage lower than the threshold will be ignored.  Default value is
-  0 (all the association will be added to the extended dataset)
-
+  * `max-rules`: (number) Maximum number of association rules
```
mmerce: integer

mmerce (approved):

PR #84, opened 2019-09-09
Stratified cross-validation

mmerce:


cross-validation/boosted-ensemble/script.whizzml
```diff
@@ -172,6 +195,29 @@
...
+                                                        "(= "
+                                                        "(f {objective-id})"
+                                                        " {{cat}})")
+                                        "name" (str (dataset "name")
+                                                    " - "
+                                                    cat)})
+                       k-folds))
+        ds-ids (for (index (range 0 (count (cat-ds-ids 0))))
+                 (create-dataset {"origin_datasets" (map (lambda (x) (x index))
+                                                         cat-ds-ids)})))
```
jaor: if i'm reading this correctly, you want to map over the transposed
`cats-ds-ids`, right? if so, a shorter way of transposing `m`
if `(apply map list m)`

```
ds-ids (map (lambda (ids) (create-dataset {"origin_datasets" ids}))
            (apply map list cats-ds-ids))
```

we could have a `transpose` defined in he stdlib... if you use this
trick here, i would define a little `_transpose` function to make
it clearer.

cross-validation/deepnet/script.whizzml
```diff
@@ -126,6 +129,23 @@
...
 
 
+;; check-minority-instances
+;;
+;; Checks whether the number of instances in the minority class is greater
+;; than the number of folds
+(define (check-minority-instances k-folds dataset-id objective-id)
+  (let (dataset (fetch dataset-id)
+        min-cat (last (dataset ["fields" objective-id "summary" "categories"]
+                               []))
```
jaor: this is going to throw an exception if there are no categories... did you mean:

```
(last (dataset ["fields" ...] []) [])
```

or

```
(last (datasest ["fields" ...] [[]])
```

cross-validation/deepnet/script.whizzml
```diff
@@ -126,6 +129,23 @@
...
+        min-inst (if (empty? min-cat)
+                     0
+                     (min-cat 1)))
```
jaor: `(min-cat 1 0)` does the same thing...

cross-validation/deepnet/script.whizzml
```diff
@@ -159,6 +179,30 @@
     (fields [objective-id "name"] false)))
 
 
+;; Stratified sampling: Keeping proportions in sampling across classes
+
+(define (stratified-k-folds dataset k-folds objective-id)
+  (let (summary (dataset ["fields" objective-id "summary" "categories"] [])
+        cat-names (map (lambda (x) (x 0)) summary)
```
jaor: nit, but `(map head summary)` is shorter

cross-validation/deepnet/script.whizzml
```diff
@@ -159,6 +179,30 @@
...
+                                                        "(= "
+                                                        "(f {objective-id})"
+                                                        " {{cat}})")
+                                        "name" (str (dataset "name")
+                                                    " - "
+                                                    cat)})
+                       k-folds))
+        ds-ids (for (index (range 0 (count (cat-ds-ids 0))))
+                 (create-dataset {"origin_datasets" (map (lambda (x) (x index))
+                                                         cat-ds-ids)})))
```
jaor: if we define a `_transpose` we can reuse it here

cross-validation/ensemble/script.whizzml
```diff
@@ -124,6 +131,22 @@
...
+;;
+;; Checks whether the number of instances in the minority class is greater
+;; than the number of folds
+(define (check-minority-instances k-folds dataset-id objective-id)
+  (let (dataset (fetch dataset-id)
+        min-cat (last (dataset ["fields" objective-id "summary" "categories"]
+                               []))
+        min-inst (if (empty? min-cat)
+                     0
+                     (min-cat 1)))
```
jaor: same as before

cross-validation/ensemble/script.whizzml
```diff
@@ -171,6 +194,28 @@
...
+                                                        "(= "
+                                                        "(f {objective-id})"
+                                                        " {{cat}})")
+                                        "name" (str (dataset "name")
+                                                    " - "
+                                                    cat)})
+                       k-folds))
+        ds-ids (for (index (range 0 (count (cat-ds-ids 0))))
+                 (create-dataset {"origin_datasets" (map (lambda (x) (x index))
+                                                         cat-ds-ids)})))
```
jaor: same as before

cross-validation/ensemble/script.whizzml
```diff
@@ -171,6 +194,28 @@
 (define (get-objective-name dataset objective-id)
   (dataset ["fields" objective-id "name"] false))
 
+;; Stratified sampling: Keeping proportions in sampling across classes
+
+(define (stratified-k-folds dataset k-folds objective-id)
+  (let (summary (dataset ["fields" objective-id "summary" "categories"] [])
+        cat-names (map (lambda (x) (x 0)) summary)
```
jaor: same as before

cross-validation/ensemble/script.whizzml
jaor: i'm stopping now :) perhaps we should consider a library?

jaor (changes requested):
a few nits... this is hard to maintain (and review) with so much code repetition, not sure if we could use a library here...

mmerce:
The cross-validation library has been separated and your suggestions added.

jaor (approved):

PR #87, opened 2019-11-04
Avoid use of non-preferred fields in AutoML

unmonoqueteclea:
Non-preferred fields are now ignored during the unsupervised models
generation step and during the creation of the extended dataset, so
we don't need to ignore them in the feature selection step because
they have already been ignored previously.

Non-preferred fields have the same behavior now as if the user had set
them in the excluded-fields configuration parameter.

### DIFF
--- a/automl/automl-library/library.whizzml
+++ b/automl/automl-library/library.whizzml
@@ -1,5 +1,5 @@
 ;;;;;;;;;;;;;;;;;;;;;;; AUTOML LIBRARY ;;;;;;;;;;;;;;;;;;;;;;;;
-;;;;;;;;;;;;;;;;;;;;; library version 0.3 ;;;;;;;;;;;;;;;;;;;;;
+;;;;;;;;;;;;;;;;;;;;; library version 0.4 ;;;;;;;;;;;;;;;;;;;;;
 ;;;;;;;;;;;;;;;;; Private functions start with _ ;;;;;;;;;;;;;;
 ;;;;;;;;;;;;;;;;;;;;; Check automl script ;;;;;;;;;;;;;;;;;;;;;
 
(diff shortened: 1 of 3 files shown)


automl/automl-script/script.whizzml
```diff
@@ -39,7 +39,9 @@
 (log-info "Obtaining unsupervised models")
 ;; Returns a list of unsupervised-models from a dataset
 (define (create-unsupervised-models dataset excluded excluded-models)
-  (let (params {"excluded_fields" (field-ids-from-names excluded dataset)}
+  (let (params {"excluded_fields"
+                (concat (non-preferred dataset)
```
mmerce: what happens if non-preferred fields are added to the excuded fields by the user?

mmerce (commented):
As a general comment, do you think that could be useful to keep the non-preferred fields as summary_fields or id_fields in some scenarios?

automl/automl-script/script.whizzml
unmonoqueteclea: Not sure, I will check....

unmonoqueteclea:
> As a general comment, do you think that could be useful to keep the non-preferred fields as summary_fields or id_fields in some scenarios?

So, do you think I should always add these fields to the final dataset? Yes, maybe you are right... I shouldn't deal with them in the same way as with excluded-fields. I will change it

unmonoqueteclea:
@mmerce, updated, now the non-preferred fields won't be removed, but they will be marked as non-preferred in all intermediate datasets.

automl/automl-library/library.whizzml
```diff
@@ -352,19 +362,21 @@
...
-(define (feature-generation-dataset ds-list name objective-id)
-  (update (create-dataset {"origin_datasets" (remove-false ds-list)
-                           "juxtapose" true
-                           "name" (str name " | extended")})
-          {"objective_field" {"id" objective-id}}))
+(define (feature-generation-dataset ds-list name objective-id non-pref)
+  (let (params {"origin_datasets" (remove-false ds-list)
+                "juxtapose" true
+                "name" (str name " | extended")})
+  (set-non-preferred (update (create-dataset params)
```
mmerce, line 377: (update ...) returns a boolean, and I think this expects a dataset-id. You can use (update-and-wait ...)

mmerce (approved):
looks good!

PR #89, opened 2019-12-09
Adding the remove-outliers script

mmerce:


unmonoqueteclea (changes requested):
This needs to be added to main repo **metadata.json** and **readme.md**

I see there is another script in the repo called `remove-anomalies` doing almost the same.
The only difference is that it receives a maximum number of features instead a maximum score.
I think we could merge both scripts and let the user choose the parameter used to filter the anomalies.

This is one example of a very easy workflow that scriptify is not able to manage (at least, the last time I tried it). Would it worth adapting scriptify for this kind of workflows or it's so complex?

mmerce:
Yes, I had a look at the script and it's a bit different because it's restricted to the anomalies detected in the anomaly detector, so its limited by its top_n parameter. However, I aggree that we could unify them, so I'll close the PR and work on a new one.

Thanks!

PR #90, opened 2020-01-15
New input to add models configuration parameters

unmonoqueteclea:
The new input map "models-configuration" will allow the user to add
configuration parameters to the created models. It allows this keys:
cluster, anomaly, association, pca, topicmodel and optiml.

The user can left it blank to use default configurations.


unmonoqueteclea:
I am also adding here a new commit to avoid the creation of unneeded resources when user provides a previous automl-execution (for test executions) Besides making the execution faster,  I think it solves a problem from a #support user.

unmonoqueteclea:
All tests are working for me now @mmerce

mmerce (approved):

PR #93, opened 2020-02-24
Adding new output to k-means minus two

mmerce:


k-means-minus-two/script.whizzml
```diff
@@ -8,51 +8,69 @@
 ;; centroid distance, the list of l anomalous points, and a list of the
 ;; similarity between sets of anomalies at each step of the algorithm.
 
+(define (headers fields)
+  (let (sorted-fields (sort-by-key "column_number" (values fields)))
+    (map (lambda (f) (f "name")) sorted-fields)))
+
+(define (make-anomalies-ds clustering)
+  (let (a-sample-id (clustering 3)
```
unmonoqueteclea, line 16: I would use a map instead of a list to avoid this magic numbers

unmonoqueteclea: We can transform it into a list at the end, to generate the final output. I think this way is easier to extend and understand

unmonoqueteclea (approved):
I checked everything and it's working for me. 
I have one suggestion and the test should be updated.

Anyways, it can be done later, we can share this now with the user that
requested the changes.

unmonoqueteclea:
@mmerce  Could you check my two little commits? I can merge if everything is ok.  Thanks!

mmerce:
looks good! I'll merge

PR #95, opened 2020-03-17
Configure AutoML complexity

unmonoqueteclea:
The new parameter `shallow_search` let the user do a more shallow (but faster) execution.

As far I have seen, accuracy results are very similar (even better sometimes) to the ones obtained with `shallow_search=False`
but executions are much faster.

I ran a benchmark and the results seems to be very good:

  | dataset    | AutoML Phi | AutoML(shallow) Phi | OptiML Phi | AutoML Time (min) | AutoML (shallow) Time (min) | OptiML Time (min) |
|-----------|----------|-----------|-----------|------------|--------------|-----------|
   | adult       |  0.6018  |     0.6089 |  0.6060 |      100 |           39 |       70 |
   | breast     |  0.9810  |     0.9427 |  0.9433 |       24 |           11 |       24 |
   | diabetes |  0.4433  |     0.5187 |  0.4491 |       29 |           12 |       27 |
   | har          |  0.9489  |     0.9301 |  0.9159 |      220 |          140 |       90 |
   | satellite   |  0.8875  |     0.8939 |  0.8770 |      156 |           50 |       61 |
   | sonar       |  0.7646 |     0.7202 |  0.7124 |      140 |           49 |       49 |
   | vehicle    |  0.6936  |     0.7411 |  0.7530 |       39 |           19 |       38 |
   | votes       |    0.9751 |     0.9751 |  0.9280 |       45 |           24 |       25 |

Some more changes:

   - Fix bug in **stratified-sampling** script
   - **Feature Selection** step is bypassed if the total number of
     fields is lower than 50 (or 30 if shallow_search is True)
   - `models-configuration` parameter appears within `configuration-params`
      map, to show a cleaner execution interface
   - `max-rules` -> `max-associations rules`
   - `pca-variance-threshold` to 0.6 by default, because we don't need
      all the Principal Components
   - Unsupervised models are created from sampled dataset if `shallow_search` is `True`
   - **RFE** is executed from sampled dataset and with a fix objective number
     of columns: 30 (if `shallow_search`)
   - `max-association-rules` refers now to the maximum number of rules
     obtained from each association discovery
   - Limited time of **OptiML**
   - Only one **association discovery** is created when `shallow_search`

### DIFF
--- a/automl/automl-script/script.whizzml
+++ b/automl/automl-script/script.whizzml
@@ -34,7 +34,7 @@
 (check-excluded-models (config "excluded-models"))
 (log-info "More configuration parameters checks")
 (check-params (config "pca-variance-threshold")
-              (config "max-associations-rules")
+              (config "max-association-rules")
               (config "validation-rate"))
 
 (define [validation-split-dataset train-split-dataset]
@@ -53,6 +53,19 @@
          {"name" (str name " - Train " (* 100 (- 1 rate)) "%")}))
       [validation-dataset (get-input "train-dataset")])))
 
+(define (sampled-ds-maybe ds-id shallow)
+  (let (columns (resource-property (wait ds-id) "columns")
+        rows (resource-property ds-id "rows")
+        name (resource-name ds-id)
+        ;; The sampled dataset should contain at least 20
+        ;; rows per dataset column
+        ratio (/ (* columns 20) rows))
+    (log-info "Sampling dataset with ratio: " ratio)
+    (if (or (> ratio 1) (not shallow))
+      ds-id
+      (create-dataset ds-id {"sample_rate" ratio
+                             "name" (str name " - sampled")}))))
+
 ;; Retrieve all non-preferred fields from all the datasets
 (define non-preferred-fields
   (remove-duplicates
@@ -66,17 +79,20 @@
 (log-featured "Feature generation")
 (log-info "Obtaining unsupervised models")
 ;; Returns a list of unsupervised-models from a dataset
-(define (create-unsupervised-models dataset excluded excluded-models)
-  (let (_ (set-non-preferred dataset non-preferred-fields)
+(define (create-unsupervised-models dataset excluded excluded-models shallow)
+  (let (dataset (sampled-ds-maybe dataset shallow)
+        _ (set-non-preferred dataset non-preferred-fields)
         params {"excluded_fields" (field-ids-from-names excluded dataset)}
         objective-id (dataset-get-objective-id dataset)
-        lev-params (merge params {"rhs_predicate" [{"field" objective-id}]
-                                  "search_strategy" "leverage"})
         lift-params (merge params {"rhs_predicate" [{"field" objective-id}]
                                    "search_strategy" "lift"})
+        ;; leverage metric is not used in shallow-search
+        lev-params (when (not shallow)
+                     (merge params {"rhs_predicate" [{"field" objective-id}]
+                                    "search_strategy" "leverage"}))
         all-models
         (map (lambda (type params*)
-               (when (create-unsupervised? type excluded-models)
+               (when (and (create-unsupervised? type excluded-models) params*)
                  (let (opt (merge params*
                                   ((config "models-configuration") type {})))
                    (create-unsupervised type dataset opt))))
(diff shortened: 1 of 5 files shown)


automl/automl-script/script.whizzml
```diff
@@ -53,6 +53,19 @@
          {"name" (str name " - Train " (* 100 (- 1 rate)) "%")}))
       [validation-dataset (get-input "train-dataset")])))
 
+(define (sampled-ds-maybe ds-id shallow)
+  (let (columns (resource-property (wait ds-id) "columns")
+        rows (resource-property ds-id "rows")
+        name (resource-name ds-id)
+        ;; The sampled dataset should contain at least 20
+        ;; rows per dataset column
+        ratio (/ (* columns 20) rows))
```
mmerce, line 62: sample_rate would be > 1 if cols are > rows/20, right?

automl/automl-script/script.whizzml
```diff
@@ -53,6 +53,19 @@
...
+    (log-info "Sampling dataset with ratio: " ratio)
+    (if (or (> ratio 1) (not shallow))
```
mmerce, line 64: ah, forget my comment. I see you just don't do shallow then

automl/automl-script/script.whizzml
```diff
@@ -94,7 +111,10 @@
         exc-models (config "excluded-models"))
     (if reuse-resources
       (get-unsupervised-models automl-execution exc-models)
-      (create-unsupervised-models train-split-dataset exc-fields exc-models))))
+      (create-unsupervised-models train-split-dataset
+                                  exc-fields
+                                  exc-models
+                                  shallow-search))))
```
mmerce, line 117: Is this defined somewhere else? or is it a typo?

mmerce: shallow-search I mean

automl/automl-script/script.whizzml
```diff
@@ -151,12 +171,31 @@
...
+                     (rfe train min-fields-shallow "" ""   "" balance)
+                     ;; We want, at least 50 fields. If the total
+                     ;; number of fields is lower, we will skip
+                     ;; feature selection stage
+                     (rfe train min-fields "" test "" balance))
         evaluations (rfe-output "evaluations" {})
-        num-features (rfe-best-num-features evaluations))
-    (rfe-output-fields train evaluations num-features)))
+        num-features (if shallow
+                       (if (> num-fields min-fields-shallow)
```
mmerce: isn't it 
(min num-fields min-fields-shallow)

automl/automl-script/script.whizzml
```diff
@@ -167,7 +206,7 @@
           [train* test*] (if (= "dataset" (resource-type valid))
                            [train valid]
                            (wait* (create-random-dataset-split train 0.8))))
-      (compute-important-fields train* test*))))
+      (compute-important-fields train* test* shallow-search))))
```
mmerce, line 207: again shallow-search or shallow....

automl/automl-script/script.whizzml
```diff
@@ -192,7 +231,8 @@
       (fusion-from-optiml
        (create-optiml-model train
                             valid
-                            (models-configuration "optiml" {}))
+                            ((config "models-configuration") "optiml" {})
+                            shallow-search)
```
mmerce, line 233: ditto

mmerce (commented):
some typos probably, but the rest looks nice!

automl/automl-script/script.whizzml
unmonoqueteclea: It's an input parameter of the script

automl/automl-script/script.whizzml
unmonoqueteclea: Yes, thanks!

automl/automl-script/script.whizzml
unmonoqueteclea: The script input is called `shallow-search`, but to avoid global variables, when I use it in a function I set it as a function argument and I name it just `shallow`

unmonoqueteclea: In the cases you pointed, as they are not inside functions, I am using the global variable name, `shallow-search`

unmonoqueteclea:
Done @mmerce

mmerce (approved):

PR #96, opened 2020-03-31
New script: Time-Aware Cross-Validation

unmonoqueteclea:
Script implementing a modified version of the Cross-Validation technique that is able to deal with **time dependent data**.

A user asked last week for this feature that will be some day included in OptiML (cc @jaor @charleslparker)


time-aware-cross-validation/readme.md
```diff
@@ -0,0 +1,77 @@
...
+* `dataset-id`: (dataset-id) Original training dataset
+* `n-splits`: (number) Number of splitting iterations in the
+  cross-validator
+* `fixed-window-size?`: (boolan) If `true`, the **Fixed Window**
+  method, with a rolling window will be used instead of the **Forward
+  Chaning** one. `false` by default
+* `model-type`: (string) Type of supervised model that shuld be
+  created. Allowed values are ``model``, ``ensemble``,
+  ``linearregression``, ``logisticregression``
+  ``deepnet`` and ``optiml``. ``ensemble`` is used by default.
```
mmerce: optiml is not a basic model

time-aware-cross-validation/script.whizzml
```diff
@@ -0,0 +1,98 @@
...
+
+(define (aggregate-evaluations evs)
+  (log-info "Creating final aggregated evaluation")
+  (create-and-wait-evaluation {"evaluations" evs}))
+
+
+;; time-aware-cross-validation
+;;
+;; Performs Cross-Validation considering temporal order.
+;; Inputs and Outputs from this function are the same as
```
mmerce: from -> for

time-aware-cross-validation/script.whizzml
```diff
@@ -0,0 +1,98 @@
...
+       (range n-splits)))
+
+;; Returns a list where each item (representing an split) contains
+;; the start and end positions of train and test rows intervals
+(define (calculate-splits rows n-splits fixed-window-size?)
+  (if fixed-window-size?
+    (calculate-splits-rolling-window rows n-splits)
+    (calculate-splits-expanding-window rows n-splits)))
+
+(define (create-cv-datasets ds-id name splits i)
```
mmerce, line 43: I'm not sure as to the method you used to compute the windows [from to] intervals. Maybe it works, but under some restrictions like rows > n-splits?

``` 
whizzml> (calculate-splits-expanding-window 3 3)
LOG info: Calculating expanding window splits...
[[[1 1] [2 2]] [[1 2] [3 2]] [[1 2] [3 3]]]
```

charleslparker:
@unmonoqueteclea I don't know if this is another script or not, but the other important case here is what I like to call "ID-based Cross Validation", where you have a column with some "group ID" field that groups rows that all came from the same "place" (like if you have multiple rows belonging to each "user").  So if you had 10 users that generated 100 points each (web site visits, say) then for five fold CV you'd hold out the points from the first two users, then the next two, then the next two and so on.

So you'd be making sub-datasets based on filtering on that "ID" field.

Does it make sense?

If you think it's a different script by all means put it off until later.

time-aware-cross-validation/script.whizzml
mmerce: And should the size of the training dataset and the test dataset be equal?

mmerce (changes requested):

time-aware-cross-validation/script.whizzml
unmonoqueteclea: Yes, I am assuming that the number of total rows is much bigger than the number of splits. In that case, I think everything is working. In the readme images you can find the expected response in a specific case, which is the same as the one obtained in the repl:

```
whizzml> (calculate-splits-expanding-window 24 5)

LOG info: Calculating expanding window splits...
[[[1 4] [5 8]]
 [[1 8] [9 12]]
 [[1 12] [13 16]]
 [[1 16] [17 20]]
 [[1 20] [21 24]]]
```
```
whizzml> (calculate-splits-rolling-window 24 5)
LOG info: Calculating rolling window splits...
[[[1 4] [5 8]]
 [[5 8] [9 12]]
 [[9 12] [13 16]]
 [[13 16] [17 20]]
 [[17 20] [21 24]]]
```

In the rolling-window case training datasets have the same length as test ones. Some of the references I used did in that way, but I don't think it's mandatory, we could create bigger training datasets and smaller test ones.

unmonoqueteclea: With the new commit I just pushed the maximum allowed value for **n-splits** is set to `0.1*rows` to avoid this cofussion

unmonoqueteclea:
> @unmonoqueteclea I don't know if this is another script or not, but the other important case here is what I like to call "ID-based Cross Validation", where you have a column with some "group ID" field that groups rows that all came from the same "place" (like if you have multiple rows belonging to each "user"). So if you had 10 users that generated 100 points each (web site visits, say) then for five fold CV you'd hold out the points from the first two users, then the next two, then the next two and so on.
> 
> So you'd be making sub-datasets based on filtering on that "ID" field.
> 
> Does it make sense?
> 
> If you think it's a different script by all means put it off until later.

It sounds great, I will do it, but I think it should be done in a different script. In this script I do groups of sequential rows, and order of these groups matters. In the one you propose, order doesn't matter, as far as instances from the same group are not separated. So, I don't think the "rolling widow" and "expanding window" methods I use here are a good approach for that case.  I think ID-Based CV is more similar to a common Cross Validation, but ensuring that groups are not separated.

charleslparker:
> It sounds great, I will do it, but I think it should be done in a different script. 

Yes, I think you understand everything precisely and a new script is probably better.

mmerce (approved):

mmerce:
I guess we can discuss about the train/test size afterwards.

PR #99, opened 2020-07-09
Adding the stratified-split script

mmerce:


stratified-split/script.whizzml
```diff
@@ -0,0 +1,82 @@
+;; Transposes the matrix: [[a b] [c d]] to [[a c] [b d]]
+
+(define (_transpose m)
+  (apply map list m))
+
+;; Creates the each dataset by sampling and using out-of-bag
```
jaor: typo (the).  also, why the blank lines between the comment of a function and its definition?

stratified-split/script.whizzml
```diff
@@ -0,0 +1,82 @@
...
+
+(define (sample-dataset ds-id rate oob)
+  (create-and-wait-dataset {"sample_rate" rate
```
jaor: if you don't wait here, split-dataset will work in parallel, possibly faster

stratified-split/script.whizzml
```diff
@@ -0,0 +1,82 @@
...
+                            "origin_dataset"  ds-id
+                            "out_of_bag" oob
+                            "seed" "bigml"}))
+
+;; Splits the original dataset in two complementary datasets using
+;; a given sample rate
+
+(define (split-dataset ds-id rate)
```
jaor: we have `create-dataset-split` in the standard library, and you can pass to it additional params (useful for setting the names of the two halves).  it also works in parallel.

stratified-split/script.whizzml
```diff
@@ -0,0 +1,82 @@
...
+         (log-info (str "Error deleting resource " id " ignored"))
+         false)))
+
+;; Check that the objective field exists and is a categorical field
+
+(define (check-cat fields objective-field)
+  (let (field (find-field fields objective-field))
+    (when (not (find-field fields objective-field))
+      (raise {"message" (str "The objective field cannot be found in this"
+                             " dataset")
```
jaor: i would perhaps print also `objective-field` in the error message, just for quick reference when seeing the error...

stratified-split/script.whizzml
```diff
@@ -0,0 +1,82 @@
...
+                       (create-dataset {"origin_dataset" ds-id
+                                        "lisp_filter" (flatline
+                                                        "(= "
+                                                        "(f {{objective-id}})"
+                                                        " {{cat}})")
+                                        "name" (str (dataset "name")
+                                                    " - "
+                                                    cat)})
+                       training-rate))
+
```
jaor: spurious blank line

stratified-split/script.whizzml
```diff
@@ -0,0 +1,82 @@
...
+        ds-ids (map (lambda (ids) (create-dataset {"origin_datasets" ids}))
+                    (_transpose cat-ds-ids) ))
+    (wait* ds-ids)
+    (map safe-delete cat-ds-ids)
+    (iterate (acc []
+              ds-id ds-ids
+              type [(str "train (" (* training-rate 100) "%)")
+                    (str "test (" (round (* 100 (- 1 training-rate))) "%)")])
+      (append acc (update ds-id {"name" (str (dataset "name") " - " type)})))))
```
jaor: if you use `create-dataset-split`, you can pass `{"name" ...}` as two extra arguments there, and avoid this update, i think.

jaor (commented):
looks mostly good: a couple of style nits, and a suggestion to use a standard function

stratified-split/script.whizzml
mmerce: Not really, I belive. The splits are done on files that contain the rows for a single class. Then the different files created in the split are joined into the final two resulting files, so if you don't do anything, the name is appended with a [merged] text

jaor:
On Thu, Jul 09 2020, mmerce wrote:

> Not really, I belive. The splits are done on files that contain the
> rows for a single class. Then the different files created in the split
> are joined into the final two resulting files, so if you don't do
> anything, the name is appended with a [merged] text

isn't the merge a create-dataset call? if so, we could specify the name
there then.

mmerce:
Requested changes added

jaor (approved):

PR #105, opened 2022-01-31
Add cluster-chronological-split whizzml script

guillemvidalfite:
Adding a new WhizzML script coming from a project into the WhizzML-examples folder.
The script is about splitting data chronologically by similarity using a cluster. The original dataset to be split is used to train a cluster and a posterior batch centroid. Dividing the data into N groups by similarity. Then the data (requirement is that the input dataset is ordered chronologically) is split in each cluster chronologically. Finally all cluster train and test portions are merged respectively, resulting into the output: a training and a test dataset resulting from this specific split.
The data split performed is suitable with time dependent data where random splits need to be avoid. In the mean time the resulting split also ensures that both training and test datasets contain a portion of data from each cluster, being more representative of all data typologies available.
The original skeleton comes from a scriptify execution but has deeply been adapted manually with the help of Mercè (thanks)

### DIFF
--- a/cluster-chronological-split/readme.md
+++ b/cluster-chronological-split/readme.md
@@ -1,0 +1,16 @@
+# Cluster 10 split
+###Split dataset chronologically into a Training and Test dataset based on a Cluster
+
+* **Inputs:** Given a dataset chronologically ordered, a set of input fields, a K number and a split proportion
+* **Outputs:** Returns a dictionary including a test and training dataset resulting from the implemented split
+
+1. From the input dataset, trains a K-means Cluster (cluster fields and K as an input)
+2. Batch centroid over the original dataset to add K different cluster labels. This way data is separated in K groups of similarity
+3. For each cluster create a new dataset
+4. Split datasets linearly (using a proportion from the input parameter) keeping the latest for test
+5. Merge all train and test datasets portions into the resulting training and test final datasets
+
+Requirements: The input dataset needs to be ordered chronologically (BigML ordering capabilities can be manually used as desired)
+
+Installation command to be executed from the local directory (bigmler needs to be previously installed):
+`bigmler whizzml --package-dir . --output-dir ~/tmp --org-project project/yourProjectId12345
--- a/cluster-chronological-split/script.whizzml
+++ b/cluster-chronological-split/script.whizzml
@@ -1,46 +1,80 @@
 ;; create cluster from inputs
-(define cluster (create-and-wait-cluster {"dataset" dataset-input
-                                           "k" num-clusters
-                                           "default_numeric_value" "mean"
-                                           "input_fields" cluster-fields-input}))
+(define cluster 
+  (create-and-wait-cluster 
+    {"dataset" dataset-input
+     "k" num-clusters
+      "default_numeric_value" default-numeric-value
+      "input_fields" cluster-fields-input}))
 
 ;; batch centroid to label input dataset
-(define batch-centroid (create-and-wait-batchcentroid {"cluster" cluster
-                                                       "dataset" dataset-input
-                                                       "all_fields" true
-                                                       "output_dataset" true}))
+(define batch-centroid 
+  (create-and-wait-batchcentroid 
+    {"cluster" cluster
+     "dataset" dataset-input
+     "all_fields" true
+     "output_dataset" true}))
 
-(define dataset-centroid ((fetch (wait batch-centroid)) ["output_dataset_resource"] false))
+(define dataset-centroid 
+  ((fetch (wait batch-centroid)) 
+    ["output_dataset_resource"] false))
 
 ;; get cluster names from resulting dataset
-(define fields (resource-fields (wait dataset-centroid)))
-(define cluster-field-id ((find-field fields "cluster") "id"))
-(define cluster-names (map head (fields [cluster-field-id "summary" "categories"])))
+(define fields 
+  (resource-fields 
+   (wait dataset-centroid)))
+
+(define cluster-field-id 
+  ((find-field fields "cluster") "id"))
+
+(define cluster-names 
+  (map head 
+    (fields [cluster-field-id "summary" 
+                              "categories"])))
 
 ;; filter centroid dataset for each cluster (returns filtered dataset list with num-clusters elements)
-(define cluster-datasets (map (lambda(x) (create-and-wait-dataset {"origin_dataset" dataset-centroid
-                                                          "lisp_filter" (flatline "(= (f \"cluster\") {{x}})")
-                                                          "excluded_fields" ["cluster"]})) cluster-names))
+(define cluster-datasets 
+  (map (lambda(x) 
+          (create-and-wait-dataset {"origin_dataset" dataset-centroid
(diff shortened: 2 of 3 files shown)


cluster-chronological-split/readme.md
```diff
@@ -0,0 +1,16 @@
+# Cluster 10 split
+###Split dataset chronologically into a Training and Test dataset based on a Cluster
+
+* **Inputs:** Given a dataset chronologically ordered, a set of input fields, a K number and a split proportion
+* **Outputs:** Returns a dictionary including a test and training dataset resulting from the implemented split
+
+1. From the input dataset, trains a K-means Cluster (cluster fields and K as an input)
+2. Batch centroid over the original dataset to add K different cluster labels. This way data is separated in K groups of similarity
+3. For each cluster create a new dataset
+4. Split datasets linearly (using a proportion from the input parameter) keeping the latest for test
```
mmerce, line 10: I was just wondering if that is a good method for a linear split... I mean, imagine that the original cluster has 2 years of data and the data for the first year is quite different from the one of the second year. Clustering should then create clusters containing only data from one of the years. If you try to do a linear split using a 80-20% per cluster, some fo the test datasets will contain data from the last part of the first year and other will contain data of the last part of the second year. Once they are merged at the end of the process, you'll end up having test data points covering more than a year of time, not the last 20% of time, as you would in a non-clustered approach... just thinking outloud about whether that can be a good train/test split in some scenarios...

cluster-chronological-split/script.whizzml
```diff
@@ -0,0 +1,46 @@
+;; create cluster from inputs
+(define cluster (create-and-wait-cluster {"dataset" dataset-input
+                                           "k" num-clusters
```
mmerce: indentation <-

cluster-chronological-split/script.whizzml
```diff
@@ -0,0 +1,46 @@
...
+                                           "default_numeric_value" "mean"
+                                           "input_fields" cluster-fields-input}))
```
mmerce: The "mean" value could be an input with a default value or a (define DEFAULT_NUMERIC_VALUE "mean") declaration at the top of the file

cluster-chronological-split/script.whizzml
```diff
@@ -0,0 +1,46 @@
...
+
+(define dataset-centroid ((fetch (wait batch-centroid)) ["output_dataset_resource"] false))
+
+;; get cluster names from resulting dataset
+(define fields (resource-fields (wait dataset-centroid)))
+(define cluster-field-id ((find-field fields "cluster") "id"))
+(define cluster-names (map head (fields [cluster-field-id "summary" "categories"])))
+
+;; filter centroid dataset for each cluster (returns filtered dataset list with num-clusters elements)
+(define cluster-datasets (map (lambda(x) (create-and-wait-dataset {"origin_dataset" dataset-centroid
```
mmerce: image

cluster-chronological-split/script.whizzml
```diff
@@ -0,0 +1,46 @@
...
+                                                          "lisp_filter" (flatline "(= (f \"cluster\") {{x}})")
+                                                          "excluded_fields" ["cluster"]})) cluster-names))
+
+;; count rows for each dataset inside a new list
+(define num-rows (map (lambda (x) ((fetch x) "rows")) cluster-datasets))
+(define num-train-rows (map (lambda (x) (round (* (/ x 100) train-percent))) num-rows))
+
+;; split cluster datasets
+(define train-datasets (map (lambda (x y)
+                                    (create-and-wait-dataset {"origin_dataset" x "range" [1, y]}))
```
mmerce: `"range" [1 y] `

no commas used in whizzml

cluster-chronological-split/script.whizzml
```diff
@@ -0,0 +1,46 @@
...
+                             cluster-datasets
+                             num-train-rows))
+
+(define test-datasets (map (lambda (x y z)
+                                    (create-and-wait-dataset {"origin_dataset" x "range" [(+ 1 y), z]}))
```
mmerce: ditto about commas

cluster-chronological-split/script.whizzml
```diff
@@ -0,0 +1,46 @@
...
+(define cluster-datasets (map (lambda(x) (create-and-wait-dataset {"origin_dataset" dataset-centroid
+                                                          "lisp_filter" (flatline "(= (f \"cluster\") {{x}})")
+                                                          "excluded_fields" ["cluster"]})) cluster-names))
+
+;; count rows for each dataset inside a new list
+(define num-rows (map (lambda (x) ((fetch x) "rows")) cluster-datasets))
+(define num-train-rows (map (lambda (x) (round (* (/ x 100) train-percent))) num-rows))
+
+;; split cluster datasets
+(define train-datasets (map (lambda (x y)
```
mmerce: try to avoid long lines as in the example above

cluster-chronological-split/script.whizzml
```diff
@@ -0,0 +1,46 @@
...
+                                    (create-and-wait-dataset {"origin_dataset" x "range" [1, y]}))
+                             cluster-datasets
+                             num-train-rows))
+
+(define test-datasets (map (lambda (x y z)
```
mmerce: same about long lines

mmerce (changes requested):
Nice script, congrats, and I agree that can be useful sometimes, when all the generated clusters have evenly-chronologically-distributed data. I just would check with Charlie in case he foresees some kind of problem in different scenarios.

cluster-chronological-split/readme.md
guillemvidalfite: Yes, I understand and am aware about this scenario. It's part of the solution and the user would need to be aware. The idea of implementing the script came from chemical plants where they have constant reactions 24/7/365 and they change the execution style now and then by modifying any of the parameters (many). It would be possible to have the scenario you mention about a cluster having only early and late data. That would not be a problem to split chronologically, it is the expected behavior from such solution. Even if training examples would be much older than test examples, by using this method we could ensure having examples of that cluster specific plant condition in both training and test datasets

cluster-chronological-split/script.whizzml
guillemvidalfite: I changed the indentation and reduced longer lines, thanks

cluster-chronological-split/script.whizzml
guillemvidalfite: added as an input

cluster-chronological-split/script.whizzml
guillemvidalfite: Left it as option 1, thanks for option 2

cluster-chronological-split/script.whizzml
guillemvidalfite: removed the commas in both places

guillemvidalfite:
All issues seem to be resolved and tested now

guillemvidalfite (commented):
I applied the changes to my local repository, figuring out how to handle this on git now :)

mmerce (approved):

PR #109, opened 2023-01-13
Adding balance-optiml to create OptiMLs using undersampling of Unbalanced Datasets

mmerce:


balanced-optiml/bal-optiml-library/library.whizzml
```diff
@@ -0,0 +1,139 @@
...
+  (try (delete id)
+       (catch e
+         (log-info (str "Error deleting resource " id " ignored:" e))
+         false)))
+
+;; Check that the objective field exists, is a categorical field and
+;; contains the rare class
+(define (_check-cat fields objective-field rare-class)
+  (let (field (find-field fields objective-field))
+    (when (not (find-field fields objective-field))
```
jaor: shouldn't this be `(when (not field))`?

balanced-optiml/bal-optiml-library/library.whizzml
```diff
@@ -0,0 +1,139 @@
...
+                             " does not contain the "
+                             rare-class
+                             " category. Please check its name.")
+              "code" 102}))))
+
+;; Function that splits a dataset in training, validation and test datasets
+(define (_create-train-validation-test ds-id train-rate seed)
+  (let (first-split (create-dataset-split ds-id train-rate seed)
+        second-split (create-dataset-split (last first-split) 0.5 seed))
+    (cons (head first-split) second-split)))
```
jaor, line 87: you could also use here  destructuring`(let ([fst snd] (create-dataset-split ds-id train-rate seed)`, but either way is fine by me.

balanced-optiml/bal-optiml-library/library.whizzml
```diff
@@ -0,0 +1,139 @@
...
+  (let (dataset (fetch ds-id)
+        ds-name (dataset "name")
+        fields (resource-fields dataset)
+        _ (_check-cat fields objective-field rare-class)
+        objective-id ((find-field fields objective-field) "id")
+        filters [(flatline "(= (f {{objective-id}}) {{rare-class}})")
+                 (flatline "(!= (f {{objective-id}}) {{rare-class}})")]
+        names [rare-class (str "not " rare-class)]
+        cat-ds-ids (iterate (acc [] expr filters name names)
+                     (append acc
```
jaor: iterate + append is just map, right? although i like map better, it's okay to keep iterate if you prefer, but then perhaps you could also use it in the computation of ds-ids below for consistency... seeing both used next to the other made me double check that iterate wasn't doing something else.

balanced-optiml/bal-optiml-library/library.whizzml
```diff
@@ -0,0 +1,139 @@
...
+                         "bigml")))
+        inner-ds (flatten cat-ds-ids)
+        ds-ids (map (lambda (ids name)
+                      (merge-datasets ids {"name" name}))
+                    (_transpose cat-ds-ids)
+                    [(str ds-name " - train")
+                     (str ds-name " - validation")
+                     (str ds-name " - test")]))
+    (wait* ds-ids)
+    (map _safe-delete inner-ds)
```
jaor, line 120: if i recall correctly, `delete*` deletes a list and also skips errors.  another option is to create the resources you want to delete with the option `{"temp" true ...}`, which marks any resource as to be deleted when the script finishes, automatically.

balanced-optiml/bal-optiml-library/library.whizzml
```diff
@@ -0,0 +1,139 @@
...
+    ds-ids))
+
+
+;; Function to generated datasets that keep all instances of the rare class
+;; while undersampling the instances in the common class to maintain a
+;; proportion with the rare class
+(define (undersample-fn ds-id number-of-samples objective-field rare-class
+         proportion)
```
jaor: nit, `proportion` aligned with `ds-id`

balanced-optiml/bal-optiml-library/library.whizzml
```diff
@@ -0,0 +1,139 @@
...
+  (let (dataset (fetch ds-id)
+        ds-name (dataset "name")
+        fields (resource-fields dataset)
+        _ (_check-cat fields objective-field rare-class)
+        objective-id ((find-field fields objective-field) "id")
```
jaor: another optional nit: you could have a function _find-objective that performs the check and save a tiny bit of repetition.

balanced-optiml/bal-optiml-library/library.whizzml
```diff
@@ -0,0 +1,139 @@
...
+        rows (dataset "rows")
+        field-summary (field-distribution (find-field fields objective-field))
+        _ (log-info field-summary)
+        rare-instances ((head (filter (lambda (f) (= (f 0) rare-class))
+                                      field-summary)) 1)
+        common-instances (- rows rare-instances)
+        balance-rate (* proportion (/ rare-instances common-instances))
```
jaor, line 141: is there any possibility (perhaps by a badly specified rare-class) that this gives a division by zero?

balanced-optiml/bal-optiml-script/metadata.json
```diff
@@ -0,0 +1,67 @@
+{
+  "name": "Balanced OptiML Script",
+  "description": "Applying undersampling techniques for datasets that contain a rare class before building OptiMLs. The result is a list of OptiMLs built on a collection of undersampled datasets and the unbalanced test data to do the final evaluation.",
+  "kind": "script",
+  "imports": ["../bal-optiml-library"],
+  "source_code": "script.whizzml",
+  "inputs": [
+    {
+      "name": "ds-id",
```
jaor: nit: for consistency with the other parameter names, none of which are abbreviated, perhaps this could be called "dataset" or "dataset-id".

balanced-optiml/bal-optiml-script/metadata.json
```diff
@@ -0,0 +1,67 @@
...
+      "type": "map",
+      "description": "OptiMLs configuration options",
+      "default": {}
+    }
+  ],
+  "outputs": [
+    {
+      "name": "undersampled-optimls",
+      "type": "list",
+      "description": "List of OptiMLs built on undersampled datasets"
```
jaor: would it make sense to also return the list of undersampled datasets, for easy reuse elsewhere?

balanced-optiml/bal-optiml-script/readme.md
```diff
@@ -0,0 +1,51 @@
...
+  split (usually the objective field).
+
+- The rare class to be balanced.
+
+- The rate of instances that you want to use for training (default is 0.8 and
+  selects 80% of the instances). Test and Validation datasets will split
+  50%-50% the hold out instances.
+
+- The number of datasets that will be created by undersampling the common
+  classes. (default = 5).
```
jaor: typo: spurious period after classes

balanced-optiml/bal-optiml-script/script.whizzml
```diff
@@ -0,0 +1,40 @@
...
+
+;; List of undersampled datasets built by keeping all the instances for the
+;; rare class and randomly sampling the same amount of instances from the
+;; rest of data using different seeds. The proportion between non-rare vs. rare
+;; instaces can also be defined.
+(define undersampled-dss (undersample-fn (head train-val-test-split)
+                                         number-of-samples
+                                         objective-field
+                                         rare-class
+                                         proportion))
```
jaor, line 24: perhaps we could check that proportion has a value within range, and number-of-samples too.  or do that in the library (i might have missed it's already done).

jaor (approved):
looks essentially good, i just added a handful of mostly cosmetic nits

balanced-optiml/bal-optiml-library/library.whizzml
mmerce: But it does not log anything in the console log, right?

mmerce: Good to know the options, though

balanced-optiml/bal-optiml-library/library.whizzml
mmerce: thanks

balanced-optiml/bal-optiml-library/library.whizzml
mmerce: ok, changing to map

balanced-optiml/bal-optiml-library/library.whizzml
mmerce: yes

balanced-optiml/bal-optiml-library/library.whizzml
mmerce: ok

balanced-optiml/bal-optiml-library/library.whizzml
mmerce: The only corner case is iff there's only instances that belong to the rare class in the dataset.  I'll add a check to avoid that case and another to prevent using datasets with less that 10 instances in the rare class to ensure that enough instances are provided in the test and validation datasets under reasonable conditions.

balanced-optiml/bal-optiml-script/metadata.json
mmerce: ok

balanced-optiml/bal-optiml-script/script.whizzml
mmerce: The proportion is used to allow multiples of the number of instances of the rare class to be the non-common instances (not only balanced datasets), so I don't see that we need to set a range for that. The number-of-samples is the number of undersampled datasets to be created, and it also can be as big as wanted, like the number of models an ensemble uses.

jaor:
but still they both must be > 0, and the proportion <= 1, no?

mmerce:
> but still they both must be > 0, and the proportion <= 1, no?

They both must be positive and number-of-samples has to be an integer.

jaor:
okay, i guess it's not specially useful to provide specific checks:
anyone using the script will know, this is for experts or people guided
by experts :) up to you!

mmerce:
> okay, i guess it's not specially useful to provide specific checks: anyone using the script will know, this is for experts or people guided by experts :) up to you!

No problem adding the changes. Done.

PR #114, opened 2025-05-02
Adding scripts to compute lags, windows aggregations, momentums and c…

mmerce:
…urvatures


time-series-utils/metadata.json
```diff
@@ -4,6 +4,8 @@
   "kind": "package",
   "components":[
     "preprocessing",
-    "epoch-ts"
+    "epoch-ts",
+    "lags-windows",
+    "momentums-curvatures"
```
jaor: i think you forgot to git add the actual scripts!

also, if you'll pardon my snobism, the plural of momentum is momenta :)

jaor (changes requested):

jaor (approved):

